PR #1, opened 2024-06-22
Add pull request template.

kirkrodrigues:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
...
+Set the PR title to a meaningful commit message in imperative form. E.g.:
+
+clp-s: Don't add implicit wildcards ('*') at the beginning and the end of a query (fixes #390).
+-->
+
+# Description
+<!-- Describe what this request will change/fix and provide any details necessary for reviewers. -->
+
+
+
```
LinZhihao-723, line 11: Do we really need three empty lines in the template?

kirkrodrigues: Yes and no:
* I like to leave an empty line for users to enter the content.
* I typically include a newline between headings and content since that's the rule I learned a long time ago based on a very relaxed Markdown spec. However, GitHub's Markdown spec seems to suggest that's not necessary and the same applies within docs, so I will drop the newlines going forward.

LinZhihao-723 (approved):
commit msg lgtm

PR #2, opened 2024-06-22
Add utility Taskfile for checksumming, text replacement, and Python venv creation.

kirkrodrigues:
# Description

Adds a Taskfile containing common utility tasks that we've been using across other repos.

To understand why we need checksumming tasks, see: https://docs.yscope.com/dev-guide/contrib-guides-taskfiles.html#sources-and-generates

# Validation performed

These tasks have been used in https://github.com/y-scope/clp

### DIFF
--- a/taskfiles/utils.yml
+++ b/taskfiles/utils.yml
@@ -1,0 +1,90 @@
+version: "3"
+
+tasks:
+  # ===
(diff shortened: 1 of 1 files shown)


LinZhihao-723 (approved):
commit msg lgtm

PR #6, opened 2024-08-19
taskfiles: Add support for excluding paths in checksumming utilities.

junhaoliao:
# Description

An issue was encountered when the checksumming utilities are used in a go-task Taskfile in clp-ffi-js https://github.com/y-scope/clp-ffi-js/pull/1 . The Taskfile downloads an SDK, calls task `compute-checksum`, and compiles some code with the SDK. However, during the compilation, there are cache files produced within the SDK directory. When developers reruns the Taskfile, checksum validation fails on the SDK directory and causes the SDK to be re-downloaded and the code to be re-compiled, which negates the effort to avoid unnecessary reprocessing.

This PR adds support for `EXCLUDE_PATHS` in checksumming utilities.

# Validation performed

1. Created an empty directory named `example`.
2. In the newly created directory, created test files and another directory with below structure:
   ```
   - example
     - Taskfile.yml (to be filled in the next step)
     - test
       - ignored
         - ignored_in_dir.txt
       - ignored_file.txt
       - important.txt
   ```
4. Filled Taskfile.yml with below content.
    ```yaml
    version: '3'
    
    includes:
      utils: "utils.yml"
    
    tasks:
      default:
        vars:
          BUILD_DIR: "test"
          OUTPUT_FILE: "checksum.txt"
        cmds:
          - task: "utils:validate-checksum"
            vars:
              CHECKSUM_FILE: "{{.OUTPUT_FILE}}"
              DATA_DIR: "{{.BUILD_DIR}}"
              EXCLUDE_PATHS: &test_excluded_paths
                - "ignored"
                - "ignored_file.txt"
          - task: "utils:compute-checksum"
            vars:
              DATA_DIR: "{{.BUILD_DIR}}"
              OUTPUT_FILE: "{{.OUTPUT_FILE}}"
              EXCLUDE_PATHS: *test_excluded_paths
          - "cat {{.OUTPUT_FILE}}"
          - "date > {{.BUILD_DIR}}/ignored/ignored_in_dir.txt"
          - "date > {{.BUILD_DIR}}/ignored_file.txt"
    ```
5. Ran `task` twice and did not observed "Files checksum.txt.tmp and checksum.txt differ".
6. Modified `example/test/important.txt`, ran task and observed "Files checksum.txt.tmp and checksum.txt differ".

### DIFF
--- a/taskfiles/utils.yml
+++ b/taskfiles/utils.yml
@@ -3,10 +3,11 @@ version: "3"
 tasks:
   # ===
   # CHECKSUM UTILS
+  # ===
+
   # @param {string} DATA_DIR The directory to compute the checksum for.
   # @param {string} OUTPUT_FILE
   # @param {[]string} [EXCLUDE_PATHS] A list of paths to exclude from the checksum.
-  # ===
   compute-checksum:
     desc: "Tries to compute a checksum for the given directory and output it to a file."
     internal: true
@@ -30,6 +31,9 @@ tasks:
     # Ignore errors so that dependent tasks don't fail
     ignore_error: true
 
+  # @param {string} DATA_DIR The directory to compute the checksum for.
+  # @param {string} OUTPUT_FILE
+  # @param {[]string} [EXCLUDE_PATHS] A list of paths to exclude from the checksum.
   validate-checksum:
     desc: "Validates the checksum of the given directory matches the checksum in the given file, or
     deletes the checksum file otherwise."
--- a/taskfiles/utils.yml
+++ b/taskfiles/utils.yml
@@ -10,6 +10,18 @@ tasks:
     silent: true
     requires:
       vars: ["DATA_DIR", "OUTPUT_FILE"]
+    vars:
+      # When the task is called with EXCLUDE_PATHS being a list, concatenate the items into a comma (",") separated
+      # string. This is needed because go-task does not support passing array variables from one task's argument into
+      # another task.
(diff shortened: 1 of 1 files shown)


taskfiles/utils.yml
```diff
@@ -19,6 +19,9 @@ tasks:
         --numeric-owner
         --owner 0
         --sort name
+        {{- range .EXCLUDE_PATHS}}
```
kirkrodrigues, line 22: Since this is an optional variable that may not be obvious to future users, we should probably document it. Initially, I was planning to document this in the docs but I think it's probably better to document it inline. So how about we document task parameters like this:

```yml
  # @param {string} DATA_DIR The directory to compute the checksum for.
  # @param {string} OUTPUT_FILE
  # @param {[]string} [EXCLUDE_PATHS] A list of paths to exclude from the checksum.
```

The docstring uses JSDoc's syntax except the types use Go's syntax. I don't think we need to document
 the return value since that's kind of nebulous for a task; and the description already has a dedicated field in the task.

kirkrodrigues (changes requested):
Nice solution. Just one suggestion.

taskfiles/utils.yml
junhaoliao: sounds good

kirkrodrigues (approved):
For the PR title, how about:

taskfiles: Add support for excluding paths in checksumming utilities.

PR #9, opened 2024-09-04
symlink-cpp-lint-configs.sh: Replace `realpath` with equivalent to support macOS.

kirkrodrigues:
# Description

macOS doesn't support `realpath --relative-to` (even with the `coreutils` package installed), so this PR replaces the instance in `symlink-cpp-lint-configs.sh` with bash variable substitutions.

# Validation performed

Validated the new script worked on both Linux and macOS in a test run of the [clp-lint](https://github.com/kirkrodrigues/clp/actions/runs/10707509343/job/29687652025) workflow on my [fork](https://github.com/kirkrodrigues/clp/tree/cpp-lint-configs) of CLP. The branch follows the [docs](https://github.com/y-scope/yscope-dev-utils/blob/main/docs/lint-tools-cpp.md) to add the C++ lint configs to the CLP repo.

### DIFF
--- a/lint-configs/symlink-cpp-lint-configs.sh
+++ b/lint-configs/symlink-cpp-lint-configs.sh
@@ -17,10 +17,14 @@ symlink_config () {
     repo_dir="$(git rev-parse --show-toplevel)"
     config_file_absolute_path="$(readlink -f "$config_file_path")"
 
-    # NOTE: This is a bit fragile since it depends on adding a trailing slash to $repo_dir. Ideally,
-    # we would use `realpath --relative-to` instead of variable substitution, but `--relative-to`
-    # isn't available on macOS.
-    repo_relative_config_file_path="${config_file_absolute_path#"${repo_dir}/"}"
+    # Ensure $repo_dir has a single trailing slash
+    repo_dir="${repo_dir%/}/"
+
+    # Get the config file's path relative to the repo root
+    # NOTE: This is a bit fragile since it depends on $repo_dir having a single trailing slash.
+    # Ideally, we would use `realpath --relative-to` instead of variable substitution, but
+    # `--relative-to` isn't available on macOS.
+    repo_relative_config_file_path="${config_file_absolute_path#"${repo_dir}"}"
 
     src_path="${repo_dir}/${repo_relative_config_file_path}"
     dst_path="${repo_dir}/$(basename "$config_file_path")"
--- a/lint-configs/symlink-cpp-lint-configs.sh
+++ b/lint-configs/symlink-cpp-lint-configs.sh
@@ -15,7 +15,12 @@ symlink_config () {
     config_file_path="$1"
 
     repo_dir="$(git rev-parse --show-toplevel)"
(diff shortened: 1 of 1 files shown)


lint-configs/symlink-cpp-lint-configs.sh
```diff
@@ -15,7 +15,12 @@ symlink_config () {
     config_file_path="$1"
 
     repo_dir="$(git rev-parse --show-toplevel)"
-    repo_relative_config_file_path="$(realpath --relative-to="$repo_dir" "$config_file_path")"
+    config_file_absolute_path="$(readlink -f "$config_file_path")"
+
+    # NOTE: This is a bit fragile since it depends on adding a trailing slash to $repo_dir. Ideally,
+    # we would use `realpath --relative-to` instead of variable substitution, but `--relative-to`
+    # isn't available on macOS.
+    repo_relative_config_file_path="${config_file_absolute_path#"${repo_dir}/"}"
```
SharafMohamed, line 23: Isn't adding the trailing slash the expected behavior? We could enforce it to be safe with:
```
repo_dir="${repo_dir%/}/"
````
and then modify repo_relative_config_file_path, src_path and dst_path to remove the added slash.

File: lint-configs/symlink-cpp-lint-configs.sh [hunk 2]
```diff
@@ -15,7 +15,12 @@ symlink_config () {
...
 
     src_path="${repo_dir}/${repo_relative_config_file_path}"
```
SharafMohamed, line 25: Isn't this always the same as `config_file_path`?

lint-configs/symlink-cpp-lint-configs.sh
SharafMohamed, line 1: Replace `return` with `exit` as we moved the code into a function with the previous PR.

lint-configs/symlink-cpp-lint-configs.sh
kirkrodrigues: Not necessarily since `config_file_path` might contain a symlink along the path. E.g.:

`a/b/c/.clang-format` where `c` is actually a symlink to `a/d`. So `config_file_path` would be `a/b/c/.clang-format` but `config_file_absolute_path` would be `a/d/.clang-format` and `src_path` would be `${repo_dir}/a/d/.clang-format`.

lint-configs/symlink-cpp-lint-configs.sh
```diff
@@ -15,7 +15,12 @@ symlink_config () {
     config_file_path="$1"
```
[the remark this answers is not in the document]
kirkrodrigues: Will do in the next PR.

lint-configs/symlink-cpp-lint-configs.sh
kirkrodrigues: Will do in the next PR.

lint-configs/symlink-cpp-lint-configs.sh
kirkrodrigues: I didn't remove the `/` from `src_path` and `dst_path` to keep them easier to read as paths (and the extra slash is harmless).

PR #10, opened 2024-09-11
symlink-cpp-lint-configs.sh: Add argument handling for `symlink_config` and exit on error.

kirkrodrigues:
# Description

This PR adds validation for the arguments passed to `symlink_config` and explicitly checks the return status so that we can exit on error.

# Validation performed

* Validated `lint-configs/symlink-cpp-lint-configs.sh` ran successfully.
* Temporarily deleted `lint-configs/.clang-format` and validated that:
  * `lint-configs/symlink-cpp-lint-configs.sh` failed because the config file didn't exist.
  * the script's return status was 1.
* Temporarily created an empty `.clang-format` and validated that:
  * `lint-configs/symlink-cpp-lint-configs.sh` failed because it already existed and wasn't a symlink to `lint-configs/.clang-format`.
  * the script's return status was 1.
* Temporarily modified `lint-configs/symlink-cpp-lint-configs.sh` to call `symlink_config` without any arguments and with multiple arguments; validated that:
  * both failed because the number of arguments was wrong.
  * the script's return status was 1.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling for configuration file symlinking, providing clearer user feedback.
	- Added validation for the number of arguments and existence of configuration files, enhancing usability.

### DIFF
--- a/lint-configs/symlink-cpp-lint-configs.sh
+++ b/lint-configs/symlink-cpp-lint-configs.sh
@@ -14,13 +14,13 @@ script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
 symlink_config () {
     if [ "$#" -ne 1 ]; then
         echo "Usage: symlink_config <config-file-path>"
-        exit 1
+        return 1
     fi
 
     config_file_path="$1"
     if [ ! -e "$config_file_path" ]; then
         echo "symlink_config: Config file doesn't exist: '$config_file_path'."
-        exit 1
+        return 1
     fi
 
     repo_dir="$(git rev-parse --show-toplevel)"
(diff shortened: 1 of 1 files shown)


lint-configs/symlink-cpp-lint-configs.sh
```diff
@@ -12,7 +12,16 @@ script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
 #
 # @param $1 Path to the config file in the repo.
 symlink_config () {
+    if [ "$#" -ne 1 ]; then
+        echo "Usage: symlink_config <config-file-path>"
```
SharafMohamed, line 16: Can redirect to >&2 to use stderr instead of stdout.

kirkrodrigues:
> This is unrelated to the current PR, but we can make `script_dir` `readonly`.

Good point. Didn't know about `readonly` and `local`. Will address in another PR.

SharafMohamed (approved):

Issue #14, opened 2024-11-06
The `validate-checksum` task fails with `task` v3.40.0 and higher.


kirkrodrigues:
The `validate-checksum` [task](https://github.com/y-scope/yscope-dev-utils/blob/fe2fb1b814636db2c171a43da0b95d92342162b0/taskfiles/utils.yml#L39) fails with `task` v3.40.0 since v3.40.0 changes the behaviour of how undefined variables are treated in the [`ref`](https://github.com/y-scope/yscope-dev-utils/blob/fe2fb1b814636db2c171a43da0b95d92342162b0/taskfiles/utils.yml#L53) attribute.

One potential fix is to use an empty list as the `ref`'s default:

```yaml
          EXCLUDE_PATHS:
            ref: "default (list) .EXCLUDE_PATHS"
```

kirkrodrigues:
Fixed by [ad576e4](https://github.com/y-scope/yscope-dev-utils/commit/ad576e43c1a43d7a6afde79fc9c3c952b7bf28bd).

kirkrodrigues:
Reported bug to task: https://github.com/go-task/task/issues/1911

PR #15, opened 2024-11-06
fix(taskfiles): Use empty list as default for optional task arguments passed via a `ref` (fixes 14).

kirkrodrigues:
# Description

As #14 describes, the behaviour of `task` has changed w.r.t. passing optional (undefined) variables via `ref`. It seems that `task` will now treat it as an empty string rather than leaving it as an undefined variable. Go is unhappy when trying to iterate over a variable which contains an empty string, but Go seems to skip iterating entirely if the variable is undefined.

I'll report this issue to the `task` devs, but in the meantime, this PR defines a default empty list for task parameters that we pass via the `ref` attribute.

# Validation performed

* Installed task v3.40.0
* Validated `task lint:check` currently fails in `clp-ffi-js`
* Applied the fix in clp-ffi-js's yscope-dev-utils submodule.
* Validated `task lint:check` now succeeds.

## Summary by CodeRabbit

- **Chores**
	- Updated the reference for the `EXCLUDE_PATHS` variable in the `validate-checksum` task for improved clarity.
	- No changes made to the core functionality of the `compute-checksum` task.

### DIFF
--- a/taskfiles/utils.yml
+++ b/taskfiles/utils.yml
@@ -50,7 +50,7 @@ tasks:
         vars:
           DATA_DIR: "{{.DATA_DIR}}"
           EXCLUDE_PATHS:
-            ref: ".EXCLUDE_PATHS"
+            ref: "default (list) .EXCLUDE_PATHS"
           OUTPUT_FILE: "{{.TMP_CHECKSUM_FILE}}"
       - defer: "rm -f '{{.TMP_CHECKSUM_FILE}}'"
       # Check that the directory exists and the checksum matches; otherwise delete the checksum file


taskfiles/utils.yml
```diff
@@ -50,7 +50,7 @@ tasks:
         vars:
           DATA_DIR: "{{.DATA_DIR}}"
           EXCLUDE_PATHS:
-            ref: ".EXCLUDE_PATHS"
+            ref: "default (list) .EXCLUDE_PATHS"
```
junhaoliao, line 53: I looked at https://taskfile.dev/reference/templating/#default-functions and believe our usage matches the explanations there.

junhaoliao (approved):

PR #16, opened 2024-11-14
feat(taskfiles): Add cmake and remote utils; Update checksum data path to an array.

davidlion:
# Description
This PR adds two new sets of utility tasks, while also updating the checksum util tasks.

#### cmake utils
This PR adds two new tasks related to building a cmake project. The first runs build and configure, while the second performs the installation. Separating these tasks allows up to build a project once, but install it multiple times to different locations.

#### remote utils
The main utility task is `fetch-src` which will download and extract a remote tar ball. The `curl` task is separated out to allow directly downloading any files.

#### checksum utils
Previously the checksum utils only operated on a single data path, which is an issue for tasks that generate files across multiple directories. This PR just changes the arguments to be an array of paths rather than a single path.

# Validation performed
Used in ffi-go irv2 beta branch.

## Summary by CodeRabbit

- **New Features**
	- Enhanced checksum computation and validation tasks now support multiple input paths using wildcard patterns.
	- Introduced new tasks for CMake operations (`cmake-config-and-build`, `cmake-install`) and remote file handling (`curl`, `download-and-extract-tar`).
- **Bug Fixes**
	- Improved error handling to prevent task failures during checksum operations.

### DIFF
--- a/taskfiles/utils.yml
+++ b/taskfiles/utils.yml
@@ -5,10 +5,10 @@ tasks:
   # CHECKSUM UTILS
   # ===
 
-  # @param {string} DATA_DIR The directory to compute the checksum for.
+  # @param {[]string} [DATA_PATHS] List of paths to compute the checksum for.
   # @param {string} OUTPUT_FILE
-  # @param {[]string} [EXCLUDE_PATHS] A list of paths, relative to `DATA_DIR`, to exclude from the
-  # checksum.
+  # @param {[]string} [EXCLUDE_PATHS] List of paths, relative to any `DATA_PATHS`, to exclude from
+  # the checksum.
   compute-checksum:
     desc: "Tries to compute a checksum for the given directory and output it to a file."
     internal: true
@@ -18,7 +18,6 @@ tasks:
     cmds:
       - >-
         tar cf -
-        --directory "{{.DATA_DIR}}"
         --group 0
         --mtime "UTC 1970-01-01"
         --numeric-owner
@@ -27,15 +26,18 @@ tasks:
         {{- range .EXCLUDE_PATHS}}
         --exclude="{{.}}"
         {{- end}}
-        . 2> /dev/null
+        {{- range .DATA_PATHS}}
+        "{{.}}"
+        {{- end}}
+        2> /dev/null
         | md5sum > {{.OUTPUT_FILE}}
     # Ignore errors so that dependent tasks don't fail
     ignore_error: true
 
-  # @param {string} DATA_DIR The directory to validate the checksum for.
+  # @param {[]string} [DATA_PATHS] List of paths to validate the checksum for.
   # @param {string} OUTPUT_FILE
-  # @param {[]string} [EXCLUDE_PATHS] A list of paths, relative to `DATA_DIR`, to exclude from the
-  # checksum.
+  # @param {[]string} [EXCLUDE_PATHS] List of paths, relative to any `DATA_PATHS`, to exclude from
+  # the checksum.
   validate-checksum:
     desc: "Validates the checksum of the given directory matches the checksum in the given file, or
     deletes the checksum file otherwise."
@@ -48,16 +50,19 @@ tasks:
     cmds:
       - task: "compute-checksum"
         vars:
-          DATA_DIR: "{{.DATA_DIR}}"
+          DATA_PATHS:
+            ref: ".DATA_PATHS"
           EXCLUDE_PATHS:
-            ref: "default (list) .EXCLUDE_PATHS"
+            ref: ".EXCLUDE_PATHS"
           OUTPUT_FILE: "{{.TMP_CHECKSUM_FILE}}"
       - defer: "rm -f '{{.TMP_CHECKSUM_FILE}}'"
       # Check that the directory exists and the checksum matches; otherwise delete the checksum file
       - >-
         (
-        test -d "{{.DATA_DIR}}"
-        && diff -q '{{.TMP_CHECKSUM_FILE}}' '{{.CHECKSUM_FILE}}' 2> /dev/null
+        {{- range .DATA_PATHS}}
+        test -e "{{.}}" &&
+        {{- end}}
+        diff -q '{{.TMP_CHECKSUM_FILE}}' '{{.CHECKSUM_FILE}}' 2> /dev/null
         ) || rm -f '{{.CHECKSUM_FILE}}'
 
   # ===
@@ -141,3 +146,159 @@ tasks:
         \( -iname "*.cpp" -o -iname "*.h" -o -iname "*.hpp" \) \
         -print0 | \
           xargs -0 clang-tidy {{.FLAGS}}
+
+  # ===
+  # CMAKE UTILS
+  # ===
+
+  # Runs cmake configure and build steps for the given source and build directories.
+  #
+  # @param {string} BUILD_DIR Cmake build directory to create.
+  # @param {string} SOURCE_DIR Project source directory containing the CMakeLists.txt file.
+  # @param {string}[optional] CHECKSUM_FILE Path to store the checksum of built files.
+  # @param {string}[optional] CMAKE_ARGS Any additional arguments to pass to cmake configure.
+  cmake-build:
+    label: "cmake-build: {{.SOURCE_DIR}} {{.BUILD_DIR}}"
+    internal: true
+    vars:
+      CHECKSUM_FILE: '{{default (printf "%s.md5" .BUILD_DIR) .CHECKSUM_FILE}}'
+      CMAKE_ARGS: "{{default nil .CMAKE_ARGS}}"
+    requires:
+      vars: ["BUILD_DIR", "SOURCE_DIR"]
+    sources:
+      - "{{.SOURCE_DIR}}/**/*"
+      - exclude: "{{.SOURCE_DIR}}/.cache/**/*"
+      - exclude: "{{.SOURCE_DIR}}/compile_commands.json"
+    generates: ["{{.CHECKSUM_FILE}}"]
+    deps:
+      - task: "validate-checksum"
+        vars:
+          CHECKSUM_FILE: "{{.CHECKSUM_FILE}}"
+          DATA_PATHS: ["{{.BUILD_DIR}}"]
+          EXCLUDE_PATHS: ["install_manifest.txt"]
+    cmds:
+      - >-
+        cmake
+        -S "{{.SOURCE_DIR}}"
+        -B "{{.BUILD_DIR}}"
+        {{.CMAKE_ARGS}}
+      - >-
+        cmake
+        --build "{{.BUILD_DIR}}"
+        --parallel
+      # This command must be last
+      - task: "compute-checksum"
+        vars:
+          DATA_PATHS: ["{{.BUILD_DIR}}"]
+          OUTPUT_FILE: "{{.CHECKSUM_FILE}}"
+          EXCLUDE_PATHS: ["install_manifest.txt"]
+
+  # Runs cmake install step for the given build directory.
+  #
+  # @param {string} BUILD_DIR Cmake build directory.
+  # @param {string} INSTALL_PREFIX Path prefix for installing the project.
+  # @param {string}[optional] CHECKSUM_FILE Path to store the checksum of installed files.
+  # @param {[]string}[optional] DATA_PATHS Paths to compute the the checksum for. Overrides using
+  # the install prefix path.
+  cmake-install:
+    label: "cmake-install: {{.BUILD_DIR}} {{.INSTALL_PREFIX}}"
+    internal: true
+    vars:
+      CHECKSUM_FILE: '{{default (printf "%s.md5" .INSTALL_PREFIX) .CHECKSUM_FILE}}'
+      # Convert the install prefix to a single element array for the default case where DATA_PATHS
+      # is not manually set.
+      _INSTALL_PREFIX_ARRAY: ["{{.INSTALL_PREFIX}}"]
+      DATA_PATHS:
+        ref: "default ._INSTALL_PREFIX_ARRAY .DATA_PATHS"
+    requires:
+      vars: ["BUILD_DIR", "INSTALL_PREFIX"]
+    sources:
+      - "{{.BUILD_DIR}}/**/*"
+      - exclude: "{{.BUILD_DIR}}/install_manifest.txt"
+    generates: ["{{.CHECKSUM_FILE}}"]
+    deps:
+      - task: "validate-checksum"
+        vars:
+          CHECKSUM_FILE: "{{.CHECKSUM_FILE}}"
+          DATA_PATHS:
+            ref: ".DATA_PATHS"
+    cmds:
+      - >-
+        cmake
+        --install "{{.BUILD_DIR}}"
+        --prefix "{{.INSTALL_PREFIX}}"
+      # This command must be last
+      - task: "compute-checksum"
+        vars:
+          DATA_PATHS:
+            ref: ".DATA_PATHS"
+          OUTPUT_FILE: "{{.CHECKSUM_FILE}}"
+
+  # ===
+  # REMOTE UTILS
+  # ===
+
+  # Runs curl to download the provided URL.
+  #
+  # @param {string} URL
+  # @param {string} URL_SHA256 Content hash to verify downloaded file against.
+  # @param {string}[optional] OUTPUT_FILE File path to store the download file.
+  curl:
+    label: "curl: {{.OUTPUT_FILE}}"
+    internal: true
+    vars:
+      OUTPUT_FILE: "{{default (base .URL) .OUTPUT_FILE}}"
+    requires:
+      vars: ["URL", "URL_SHA256"]
+    generates:
+      - "{{.OUTPUT_FILE}}"
+    status:
+      - >-
+        diff
+        <(echo "{{.URL_SHA256}}")
+        <(openssl dgst -sha256 "{{.OUTPUT_FILE}}"
+        | awk '{print $2}')
+    cmds:
+      - |-
+        mkdir -p "{{dir .OUTPUT_FILE}}"
+        curl -L "{{.URL}}" -o "{{.OUTPUT_FILE}}"
+
+  # Runs curl to download the provided URL and tar to extract its contents.
+  #
+  # @param {string} OUTPUT_DIR Path to extract downloaded tar file contents to.
+  # @param {string} URL
+  # @param {string} URL_SHA256 Content hash to verify downloaded tar file against.
+  # @param {string}[optional] CHECKSUM_FILE File path to store the checksum of downloaded tar file.
+  # @param {int}[optional] STRIP Number of leading components to strip from file names for tar.
+  # @param {string}[optional] TAR_FILE File path to store the downloaded tar file.
+  fetch-src:
+    label: "fetch-src: {{.OUTPUT_DIR}}"
+    internal: true
+    vars:
+      CHECKSUM_FILE: '{{default (printf "%s.md5" .OUTPUT_DIR) .CHECKSUM_FILE}}'
+      STRIP: "{{default 1 .STRIP}}"
+      TAR_FILE: '{{default (printf "%s.tar.gz" .OUTPUT_DIR) .TAR_FILE}}'
+    requires:
+      vars: ["OUTPUT_DIR", "URL", "URL_SHA256"]
+    sources: ["{{.TASKFILE}}"]
+    generates: ["{{.CHECKSUM_FILE}}", "{{.TAR_FILE}}"]
+    deps:
+      - task: "curl"
+        vars:
+          URL: "{{.URL}}"
+          URL_SHA256: "{{.URL_SHA256}}"
+          OUTPUT_FILE: "{{.TAR_FILE}}"
+      - task: "validate-checksum"
+        vars:
+          CHECKSUM_FILE: "{{.CHECKSUM_FILE}}"
+          DATA_PATHS: ["{{.OUTPUT_DIR}}"]
+    cmds:
+      - |-
+        rm -rf "{{.OUTPUT_DIR}}"
+        mkdir -p "{{.OUTPUT_DIR}}"
+        tar -x --strip-components="{{.STRIP}}" -C "{{.OUTPUT_DIR}}" -f "{{.TAR_FILE}}"
+      # This command must be last
+      - task: "compute-checksum"
+        vars:
+          DATA_PATHS: ["{{.OUTPUT_DIR}}"]
+          OUTPUT_FILE: "{{.CHECKSUM_FILE}}"
--- a/taskfiles/utils.yml
+++ b/taskfiles/utils.yml
@@ -20,13 +20,14 @@ tasks:
     requires:
       vars: ["INCLUDE_PATTERNS", "OUTPUT_FILE"]
     cmds:
-      # We explicitly set --no-anchored and --wildcards to ensure the behaviour for both exclusion
-      # and inclusion matches, as they are only the default for exclusion.
-      # We can quote the exclude patterns for tar because the patterns themselves are passed to tar
-      # as arguments and evaluated by tar during execution. However, for tar --create the input
-      # patterns cannot be quoted as they are evaluated by the shell and the results are passed to
-      # tar as arguments. If the input patterns are passed to tar with quotes the pattern will not
-      # be evaluated and will be treated literally.
+      # We explicitly set `--no-anchored` and `--wildcards` to make the inclusion behaviour match
+      # the default exclusion behaviour.
+      #
+      # We can quote the exclude patterns for `tar` because the patterns themselves are passed to
+      # `tar` as arguments and evaluated by `tar` during execution. However, for `tar --create`, the
+      # input patterns cannot be quoted since they're evaluated by the shell and the results are
+      # passed to `tar` as arguments. If the input patterns are passed to `tar` with quotes, the
+      # pattern won't be evaluated and will instead be treated literally.
       - >-
         tar
         --create
@@ -49,10 +50,10 @@ tasks:
     # Ignore errors so that dependent tasks don't fail
     ignore_error: true
 
+  # @param {string} CHECKSUM_FILE
   # @param {string[]} INCLUDE_PATTERNS Path wildcard patterns to validate the checksum for.
-  # @param {string} OUTPUT_FILE
-  # @param {string[]} [EXCLUDE_PATTERNS] Path wildcard patterns, relative to any `INCLUDE_PATTERNS`, to
-  # exclude from the checksum.
+  # @param {string[]} [EXCLUDE_PATTERNS] Path wildcard patterns, relative to any `INCLUDE_PATTERNS`,
+  # to exclude from the checksum.
   validate-checksum:
     desc: "Validates the checksum of the given directory matches the checksum in the given file, or
     deletes the checksum file otherwise."
@@ -75,7 +76,7 @@ tasks:
       # Check that all paths exist and the checksum matches; otherwise delete the checksum file.
       - |-
         (
-          {{- range .DATA_PATTERNS}}
+          {{- range .INCLUDE_PATTERNS}}
           for path in {{.}}; do
             test -e "$path"
           done
@@ -174,16 +175,16 @@ tasks:
   # @param {string} BUILD_DIR CMake build directory to create.
   # @param {string} SOURCE_DIR Project source directory containing the CMakeLists.txt file.
   # @param {int} [JOBS] The maximum number of concurrent processes to use when building. If
-  # omitted the native build tool's default number is used. See man cmake.
+  # omitted, the native build tool's default number is used. See `man cmake`.
   # @param {string} [CONF_ARGS] Any additional arguments to pass to CMake's configure step.
   cmake-config-and-build:
     internal: true
     label: "{{.TASK}}-{{.SOURCE_DIR}}-{{.BUILD_DIR}}"
     vars:
-      JOBS: >-
-        {{default "" .JOBS}}
       CONF_ARGS: >-
         {{default "" .CONF_ARGS}}
+      JOBS: >-
+        {{default "" .JOBS}}
     requires:
       vars: ["BUILD_DIR", "SOURCE_DIR"]
     cmds:
@@ -276,15 +277,20 @@ tasks:
     internal: true
     label: "{{.TASK}}-{{.OUTPUT_DIR}}"
     vars:
+      # Paths
       CHECKSUM_FILE: >-
         {{default (printf "%s.md5" .OUTPUT_DIR) .CHECKSUM_FILE}}
-      INCLUDE_PATTERNS:
-        ref: "default (list) .INCLUDE_PATTERNS"
-      EXCLUDE_PATTERNS:
-        ref: "default (list) .EXCLUDE_PATTERNS"
-      NUM_COMPONENTS_TO_STRIP: "{{default 1 .NUM_COMPONENTS_TO_STRIP}}"
       TAR_FILE: >-
         {{default (printf "%s.tar.gz" .OUTPUT_DIR) .TAR_FILE}}
+      
+      # Path patterns
+      EXCLUDE_PATTERNS:
+        ref: "default (list) .EXCLUDE_PATTERNS"
+      INCLUDE_PATTERNS:
+        ref: "default (list) .INCLUDE_PATTERNS"
+      
+      # Misc
+      NUM_COMPONENTS_TO_STRIP: "{{default 1 .NUM_COMPONENTS_TO_STRIP}}"
     requires:
       vars: ["FILE_SHA256", "OUTPUT_DIR", "URL"]
     sources: ["{{.TASKFILE}}", "{{.TAR_FILE}}"]
@@ -303,8 +309,8 @@ tasks:
       - |-
         rm --force --recursive "{{.OUTPUT_DIR}}"
         mkdir --parents  "{{.OUTPUT_DIR}}"
-      # Unlike tar --create, in --extract both the exclude and include patterns can be quoted as
-      # they are both evaluted by tar.
+      # Unlike for `tar --create`, in `--extract` both the exclude and include patterns can be
+      # quoted since they're both evaluated by `tar`.
       - >-
         tar
         --extract
--- a/taskfiles/utils.yml
+++ b/taskfiles/utils.yml
@@ -56,7 +56,7 @@ tasks:
             ref: ".EXCLUDE_PATHS"
           OUTPUT_FILE: "{{.TMP_CHECKSUM_FILE}}"
       - defer: "rm -f '{{.TMP_CHECKSUM_FILE}}'"
-      # Check that the directory exists and the checksum matches; otherwise delete the checksum file
+      # Check that the path exists and the checksum matches; otherwise delete the checksum file
       - >-
         (
         {{- range .DATA_PATHS}}
--- a/taskfiles/utils.yml
+++ b/taskfiles/utils.yml
@@ -53,7 +53,7 @@ tasks:
           DATA_PATHS:
             ref: ".DATA_PATHS"
           EXCLUDE_PATHS:
-            ref: ".EXCLUDE_PATHS"
+            ref: "default (list) .EXCLUDE_PATHS"
           OUTPUT_FILE: "{{.TMP_CHECKSUM_FILE}}"
       - defer: "rm -f '{{.TMP_CHECKSUM_FILE}}'"
       # Check that the path exists and the checksum matches; otherwise delete the checksum file
--- a/taskfiles/utils.yml
+++ b/taskfiles/utils.yml
@@ -5,9 +5,9 @@ tasks:
   # CHECKSUM UTILS
   # ===
 
-  # @param {[]string} [DATA_PATHS] List of paths to compute the checksum for.
+  # @param {string[]} DATA_PATHS List of paths to compute the checksum for.
   # @param {string} OUTPUT_FILE
-  # @param {[]string} [EXCLUDE_PATHS] List of paths, relative to any `DATA_PATHS`, to exclude from
+  # @param {string[]} [EXCLUDE_PATHS] List of paths, relative to any `DATA_PATHS`, to exclude from
   # the checksum.
   compute-checksum:
     desc: "Tries to compute a checksum for the given directory and output it to a file."
@@ -34,9 +34,9 @@ tasks:
     # Ignore errors so that dependent tasks don't fail
     ignore_error: true
 
-  # @param {[]string} [DATA_PATHS] List of paths to validate the checksum for.
+  # @param {string[]} DATA_PATHS List of paths to validate the checksum for.
   # @param {string} OUTPUT_FILE
-  # @param {[]string} [EXCLUDE_PATHS] List of paths, relative to any `DATA_PATHS`, to exclude from
+  # @param {string[]} [EXCLUDE_PATHS] List of paths, relative to any `DATA_PATHS`, to exclude from
   # the checksum.
   validate-checksum:
     desc: "Validates the checksum of the given directory matches the checksum in the given file, or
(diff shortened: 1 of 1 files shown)


taskfiles/utils.yml
```diff
@@ -48,16 +50,19 @@ tasks:
     cmds:
       - task: "compute-checksum"
         vars:
-          DATA_DIR: "{{.DATA_DIR}}"
+          DATA_PATHS:
+            ref: ".DATA_PATHS"
           EXCLUDE_PATHS:
-            ref: "default (list) .EXCLUDE_PATHS"
+            ref: ".EXCLUDE_PATHS"
```
junhaoliao, line 56: I believe this default fallback is needed for Task `3.40.0`. Was the removal intentional?
```suggestion
            ref: "default (list) .EXCLUDE_PATHS"
```

taskfiles/utils.yml
```diff
@@ -48,16 +50,19 @@ tasks:
...
           OUTPUT_FILE: "{{.TMP_CHECKSUM_FILE}}"
       - defer: "rm -f '{{.TMP_CHECKSUM_FILE}}'"
       # Check that the directory exists and the checksum matches; otherwise delete the checksum file
```
junhaoliao, line 59:
```suggestion
      # Check that the path exists and the checksum matches; otherwise delete the checksum file
```

taskfiles/utils.yml
```diff
@@ -141,3 +146,159 @@ tasks:
...
+  # @param {string} BUILD_DIR Cmake build directory to create.
+  # @param {string} SOURCE_DIR Project source directory containing the CMakeLists.txt file.
+  # @param {string}[optional] CHECKSUM_FILE Path to store the checksum of built files.
+  # @param {string}[optional] CMAKE_ARGS Any additional arguments to pass to cmake configure.
+  cmake-build:
+    label: "cmake-build: {{.SOURCE_DIR}} {{.BUILD_DIR}}"
+    internal: true
+    vars:
+      CHECKSUM_FILE: '{{default (printf "%s.md5" .BUILD_DIR) .CHECKSUM_FILE}}'
+      CMAKE_ARGS: "{{default nil .CMAKE_ARGS}}"
```
junhaoliao, line 165: Would it be better to fallback to `""`?

taskfiles/utils.yml
davidlion: No sorry. I think I might need to add this in other places too.

taskfiles/utils.yml
davidlion: Sure we can. I don't know if it makes a difference, but I don't think it can hurt either.

kirkrodrigues (changes requested):
Some quick high-level comments before a detailed review:

* The Rabbit has some suggestions that I think are valid.
* Can `fetch-src` replace CLP's [download-and-extract-tar](https://github.com/y-scope/clp/blob/53c4f52b3ae2ef0e990a8d66a7be87fd1e85cfbe/Taskfile.yml#L329)? If so, can we rename `fetch-src` to `download-and-extract-tar`?
* What docstring format are you basing things on?

taskfiles/utils.yml
```diff
@@ -141,3 +146,159 @@ tasks:
...
+          OUTPUT_FILE: "{{.TAR_FILE}}"
+      - task: "validate-checksum"
+        vars:
+          CHECKSUM_FILE: "{{.CHECKSUM_FILE}}"
+          DATA_PATHS: ["{{.OUTPUT_DIR}}"]
+    cmds:
+      - |-
+        rm -rf "{{.OUTPUT_DIR}}"
+        mkdir -p "{{.OUTPUT_DIR}}"
+        tar -x --strip-components="{{.STRIP}}" -C "{{.OUTPUT_DIR}}" -f "{{.TAR_FILE}}"
```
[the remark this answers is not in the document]
davidlion: Task will already fail if any command fails.

davidlion:
> Some quick high-level comments before a detailed review:
> 
>     * The Rabbit has some suggestions that I think are valid.
> 
>     * Can `fetch-src` replace CLP's [download-and-extract-tar](https://github.com/y-scope/clp/blob/53c4f52b3ae2ef0e990a8d66a7be87fd1e85cfbe/Taskfile.yml#L329)? If so, can we rename `fetch-src` to `download-and-extract-tar`?
> 
>     * What docstring format are you basing things on?

New commits have added in Rabbit's suggestions, updated `fetch-src` to provide functionality of `download-and-extract-tar` (and renamed it), and updated docstrings based on disucssion.

taskfiles/utils.yml
```diff
@@ -5,20 +5,19 @@ tasks:
...
-    desc: "Tries to compute a checksum for the given directory and output it to a file."
+    desc: "Tries to compute a checksum for the given paths and output it to a file."
     internal: true
     silent: true
     requires:
-      vars: ["DATA_DIR", "OUTPUT_FILE"]
+      vars: ["DATA_PATHS", "OUTPUT_FILE"]
     cmds:
       - >-
         tar cf -
```
kirkrodrigues, line 20: Should we add `--wildcards` and `--no-anchored` like you do for the `download-and-extract-tar` task?

kirkrodrigues:
Review incoming.

taskfiles/utils.yml
```diff
@@ -27,15 +26,18 @@ tasks:
...
         | md5sum > {{.OUTPUT_FILE}}
     # Ignore errors so that dependent tasks don't fail
     ignore_error: true
 
-  # @param {string} DATA_DIR The directory to validate the checksum for.
+  # @param {string[]} DATA_PATHS List of paths to validate the checksum for.
   # @param {string} OUTPUT_FILE
-  # @param {[]string} [EXCLUDE_PATHS] A list of paths, relative to `DATA_DIR`, to exclude from the
-  # checksum.
+  # @param {string[]} [EXCLUDE_PATHS] List of paths, relative to any `DATA_PATHS`, to exclude from
```
kirkrodrigues, line 39: Ditto suggestions from `compute-checksum`'s docstring.

taskfiles/utils.yml
```diff
@@ -44,21 +46,24 @@ tasks:
...
         vars:
-          DATA_DIR: "{{.DATA_DIR}}"
+          DATA_PATHS:
+            ref: ".DATA_PATHS"
           EXCLUDE_PATHS:
             ref: "default (list) .EXCLUDE_PATHS"
           OUTPUT_FILE: "{{.TMP_CHECKSUM_FILE}}"
       - defer: "rm -f '{{.TMP_CHECKSUM_FILE}}'"
-      # Check that the directory exists and the checksum matches; otherwise delete the checksum file
+      # Check that the path exists and the checksum matches; otherwise delete the checksum file
```
kirkrodrigues, line 59:
```suggestion
      # Check that the paths exist and the checksum matches; otherwise delete the checksum file.
```

taskfiles/utils.yml
```diff
@@ -141,3 +146,194 @@ tasks:
         \( -iname "*.cpp" -o -iname "*.h" -o -iname "*.hpp" \) \
         -print0 | \
           xargs -0 clang-tidy {{.FLAGS}}
+
+  # ===
+  # CMAKE UTILS
+  # ===
+
+  # Runs cmake configure and build steps for the given source and build directories.
```
kirkrodrigues, line 154:
```suggestion
  # Runs CMake's configure and build steps for the given source and build directories.
```

taskfiles/utils.yml
```diff
@@ -141,3 +146,194 @@ tasks:
...
+  #
+  # @param {string} BUILD_DIR Cmake build directory to create.
```
kirkrodrigues, line 156:
```suggestion
  # @param {string} BUILD_DIR CMake build directory to create.
```

taskfiles/utils.yml
```diff
@@ -141,3 +146,194 @@ tasks:
...
+  # @param {string} SOURCE_DIR Project source directory containing the CMakeLists.txt file.
+  # @param {string={{.BUILD_DIR}}.md5} [CHECKSUM_FILE] Path to store the checksum of built files.
+  # @param {string=""} [CMAKE_ARGS] Any additional arguments to pass to cmake configure.
```
kirkrodrigues, line 159:
```suggestion
  # @param {string=""} [CMAKE_ARGS] Any additional arguments to pass to CMake's configure step.
```

taskfiles/utils.yml
```diff
@@ -141,3 +146,194 @@ tasks:
...
+  cmake-build:
```
kirkrodrigues, line 160: Can we rename this to `cmake-config-and-build`?

taskfiles/utils.yml
```diff
@@ -141,3 +146,194 @@ tasks:
...
+    internal: true
+    vars:
+      CHECKSUM_FILE: >-
+        {{default (printf "%s.md5" .BUILD_DIR) .CHECKSUM_FILE}}
+      CMAKE_ARGS: >-
+        {{default "" .CMAKE_ARGS}}
+    requires:
+      vars: ["BUILD_DIR", "SOURCE_DIR"]
+    sources:
+      - "{{.SOURCE_DIR}}/**/*"
```
kirkrodrigues, line 171: Note: Currently, this won't work for clp since the top-level CMakeLists.txt is in a directory with a lot of other files, but I guess the best solution is move all the CMakeLists and source files into a dedicated directory.

taskfiles/utils.yml
```diff
@@ -141,3 +146,194 @@ tasks:
...
+      - exclude: "{{.SOURCE_DIR}}/.cache/**/*"
```
kirkrodrigues, line 172: What is this and what's generating it?

taskfiles/utils.yml
```diff
@@ -141,3 +146,194 @@ tasks:
...
+      - exclude: "{{.SOURCE_DIR}}/compile_commands.json"
```
kirkrodrigues, line 173: Isn't this generated in the build directory?

taskfiles/utils.yml
```diff
@@ -141,3 +146,194 @@ tasks:
...
+      - >-
+        cmake
+        --build "{{.BUILD_DIR}}"
+        --parallel
+      # This command must be last
+      - task: "compute-checksum"
+        vars:
+          DATA_PATHS: ["{{.BUILD_DIR}}"]
+          OUTPUT_FILE: "{{.CHECKSUM_FILE}}"
+          EXCLUDE_PATHS: ["install_manifest.txt"]
```
kirkrodrigues, line 196: Nit:
```suggestion
          EXCLUDE_PATHS: ["install_manifest.txt"]
          OUTPUT_FILE: "{{.CHECKSUM_FILE}}"
```

taskfiles/utils.yml
```diff
@@ -141,3 +146,194 @@ tasks:
...
+
+  # Runs cmake install step for the given build directory.
```
kirkrodrigues, line 198:
```suggestion
  # Runs the CMake install step for the given build directory.
```

taskfiles/utils.yml
```diff
@@ -141,3 +146,194 @@ tasks:
...
+  #
+  # @param {string} BUILD_DIR Cmake build directory.
```
kirkrodrigues, line 200:
```suggestion
  # @param {string} BUILD_DIR CMake build directory.
```

taskfiles/utils.yml
```diff
@@ -141,3 +146,194 @@ tasks:
...
+  # @param {string} INSTALL_PREFIX Path prefix for installing the project.
```
kirkrodrigues, line 201:
```suggestion
  # @param {string} INSTALL_PREFIX Path prefix of where the project should be installed.
```

taskfiles/utils.yml
```diff
@@ -141,3 +146,194 @@ tasks:
...
+  # @param {string={{.INSTALL_PREFIX}}.md5} [CHECKSUM_FILE] Path to store the checksum of installed
+  # files.
+  # @param {string[]=[{{.INSTALL_PREFIX}}]} [DATA_PATHS] Paths to compute the the checksum for.
```
kirkrodrigues, line 204:
```suggestion
  # @param {string[]=[{{.INSTALL_PREFIX}}]} [DATA_PATHS] Paths to compute the checksum for.
```

taskfiles/utils.yml
```diff
@@ -141,3 +146,194 @@ tasks:
...
+    label: "{{.TASK}}-{{.BUILD_DIR}}-{{.INSTALL_PREFIX}}"
+    internal: true
+    vars:
+      CHECKSUM_FILE: >-
+        {{default (printf "%s.md5" .INSTALL_PREFIX) .CHECKSUM_FILE}}
+      # Convert the install prefix to a single element array in the default case where DATA_PATHS
+      # has not been set.
+      _INSTALL_PREFIX_ARRAY: ["{{.INSTALL_PREFIX}}"]
+      DATA_PATHS:
+        ref: "default ._INSTALL_PREFIX_ARRAY .DATA_PATHS"
```
kirkrodrigues, line 215:
```suggestion
      DATA_PATHS:
        ref: "default (list .INSTALL_PREFIX) .DATA_PATHS"
```

I think that's equivalent?

taskfiles/utils.yml
```diff
@@ -141,3 +146,194 @@ tasks:
...
+        vars:
+          DATA_PATHS:
+            ref: ".DATA_PATHS"
+          OUTPUT_FILE: "{{.CHECKSUM_FILE}}"
+
+  # ===
+  # REMOTE UTILS
+  # ===
+
+  # Runs curl to download the provided URL.
```
kirkrodrigues, line 244:
```suggestion
  # Runs curl to download a file from the given URL.
```

taskfiles/utils.yml
```diff
@@ -141,3 +146,194 @@ tasks:
...
+  #
+  # @param {string} URL
+  # @param {string} URL_SHA256 Content hash to verify downloaded file against.
```
kirkrodrigues, line 247: How about `FILE_SHA256`?

taskfiles/utils.yml
```diff
@@ -141,3 +146,194 @@ tasks:
...
+  # @param {string={{(base .URL)}}} [OUTPUT_FILE] File path to store the download file.
```
kirkrodrigues, line 248:
```suggestion
  # @param {string={{(base .URL)}}} [OUTPUT_FILE] Path where the file should be stored.
```

taskfiles/utils.yml
```diff
@@ -141,3 +146,194 @@ tasks:
...
+  curl:
+    label: "{{.TASK}}-{{.OUTPUT_FILE}}"
+    internal: true
+    vars:
+      OUTPUT_FILE: "{{default (base .URL) .OUTPUT_FILE}}"
+    requires:
+      vars: ["URL", "URL_SHA256"]
+    generates:
+      - "{{.OUTPUT_FILE}}"
```
kirkrodrigues, line 257:
```suggestion
    generates: ["{{.OUTPUT_FILE}}"]
```

taskfiles/utils.yml
```diff
@@ -141,3 +146,194 @@ tasks:
...
+        mkdir -p "{{dir .OUTPUT_FILE}}"
+        max_attempts=3
+        attempt=1
+        while [ $attempt -le $max_attempts ]; do
+          if curl --fail --location --show-error --connect-timeout 10 --max-time 300 \
+              --location "{{.URL}}" --output "{{.OUTPUT_FILE}}";
+          then
+            break
+          fi
+          echo "Attempt $attempt failed. Retrying..."
```
kirkrodrigues, line 275: Won't this print too many times? Another way to write it is to move the loop breaking condition down here. (Fwiw, I feel like I've seen this code before and commented on the same issue. I guess it's from a StackOverflow post, lol.)

taskfiles/utils.yml
```diff
@@ -141,3 +146,194 @@ tasks:
...
+          attempt=$((attempt + 1))
+          sleep 5
+        done
+        if [ $attempt -gt $max_attempts ]; then
+          echo "Failed to download after $max_attempts attempts"
+          exit 1
+        fi
+
+  # Runs curl to download the provided URL and tar to extract its contents.
```
kirkrodrigues, line 284:
```suggestion
  # Runs curl to download the tarball from the given URL and extracts its contents.
```

taskfiles/utils.yml
```diff
@@ -141,3 +146,194 @@ tasks:
...
+  #
+  # @param {string} OUTPUT_DIR Path to extract downloaded tar file contents to.
+  # @param {string} URL
+  # @param {string} URL_SHA256 Content hash to verify downloaded tar file against.
+  # @param {string={{.OUTPUT_DIR}}.md5} [CHECKSUM_FILE] File path to store the checksum of
```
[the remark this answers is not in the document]
davidlion: What do you think about `FILE_SHA256` to match the other task?

taskfiles/utils.yml
davidlion: I'm just going to remove the log. The final log covers what happened in a failure anyway.

>Another way to write it is to move the loop breaking condition down here.
Fwiw, I'm pretty sure we'd need another condition to know whether we will retry again or not.

>(Fwiw, I feel like I've seen this code before and commented on the same issue. I guess it's from a StackOverflow post, lol.)
ya... it is from the rabbit

taskfiles/utils.yml
davidlion: I might be missing something, but this is what I've been going off of:

Without allowing the caller to specify the install prefix, we would always install to the system path by default. Installing to the system path could lead to conflicts or permission issues. The user could manually sets the `CMAKE_INSTALL_PREFIX` environment variable before running `task` but this would blindly apply to all `cmake-install` tasks.

For example: in ffi-go when we build the dependencies, I install them into the build directory (rather than the system) to ensure there are no conflicts and easy clean up.

ref: https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html

taskfiles/utils.yml
davidlion: Sorry, I'll remove this. It is old code from when I was figuring out task. Iirc it is from my vim linting setup I think.

taskfiles/utils.yml
davidlion: Sorry, I'll remove this. It is old code from when I was figuring out task.

You're right, but this line was aimed at a symlink related to my vim linting setup.

taskfiles/utils.yml
```diff
@@ -5,20 +5,19 @@ tasks:
   # CHECKSUM UTILS
   # ===
 
-  # @param {string} DATA_DIR The directory to compute the checksum for.
+  # @param {string[]} DATA_PATHS List of paths to compute the checksum for.
   # @param {string} OUTPUT_FILE
-  # @param {[]string} [EXCLUDE_PATHS] A list of paths, relative to `DATA_DIR`, to exclude from the
-  # checksum.
+  # @param {string[]} [EXCLUDE_PATHS] List of paths, relative to any `DATA_PATHS`, to exclude from
+  # the checksum.
```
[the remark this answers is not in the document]
davidlion: Want to change `DATA_PATHS` to `DATA_PATTERNS` to match?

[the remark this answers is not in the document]
kirkrodrigues: Sure

taskfiles/utils.yml
kirkrodrigues: > What do you think about changing SOURCE_DIR to SOURCE_PATHS and default CMAKELISTS_PATH to SOURCE_PATHS[0]?

Hm, would that work with `sources`? Like is there a way to pass a list variable to `sources`?

taskfiles/utils.yml
kirkrodrigues: Oh I'm fine with `INSTALL_PREFIX`---I mean what do we need `DATA_PATHS` for?

davidlion: Sorry too asleep when I responded lol.

We only want to compute the checksum for files we install, but a user might want to install to a file with other files. 
For example, say install prefix is `/usr/local` and we're installing library `foo`. We could end up with:
```
/usr/local/lib/bar
/usr/local/lib/foo
```

In a perfect world we could know what was installed from CMake (e.g. `lib/foo`), but that sounds hard :(.

Alternatively, I guess we could just always install and don't bother computing a checksum or using generates?

taskfiles/utils.yml
davidlion: Hmm good point... I'll look into it.

taskfiles/utils.yml
```diff
@@ -5,37 +5,41 @@ tasks:
...
         --sort name
-        {{- range .EXCLUDE_PATHS}}
+        --no-anchored
+        --wildcards
+        {{- range .EXCLUDE_PATTERNS}}
         --exclude="{{.}}"
         {{- end}}
-        . 2> /dev/null
+        {{- range .DATA_PATTERNS}}
+        "{{.}}"
```
kirkrodrigues, line 32: Will quoting a wildcard pattern work?

taskfiles/utils.yml
```diff
@@ -44,21 +48,24 @@ tasks:
...
       - defer: "rm -f '{{.TMP_CHECKSUM_FILE}}'"
-      # Check that the directory exists and the checksum matches; otherwise delete the checksum file
+      # Check that the paths exist and the checksum matches; otherwise delete the checksum file.
       - >-
         (
-        test -d "{{.DATA_DIR}}"
-        && diff -q '{{.TMP_CHECKSUM_FILE}}' '{{.CHECKSUM_FILE}}' 2> /dev/null
-        ) || rm -f '{{.CHECKSUM_FILE}}'
+        {{- range .DATA_PATTERNS}}
+        test -e "{{.}}" &&
```
kirkrodrigues, line 65: (Ignoring the quotes) I probably should've anticipated this, but this won't work for patterns since if a wildcard path glob returns multiple files, `test -e` will fail because it only expects one argument.

taskfiles/utils.yml
```diff
@@ -141,3 +148,158 @@ tasks:
...
+        cmake
+        -S "{{.SOURCE_DIR}}"
+        -B "{{.BUILD_DIR}}"
+        {{.CMAKE_ARGS}}
+      - >-
+        cmake
+        --build "{{.BUILD_DIR}}"
+        --parallel
+
+  # Unconditionally runs the CMake install step for the given build directory.
```
kirkrodrigues, line 180:
```suggestion
  # Runs the CMake install step for the given build directory.
```

Unless you think the unconditionally is important?

taskfiles/utils.yml
```diff
@@ -141,3 +148,158 @@ tasks:
...
+        --prefix "{{.INSTALL_PREFIX}}"
+
+  # ===
+  # REMOTE UTILS
+  # ===
+
+  # Runs curl to download a file from the given URL.
+  #
+  # @param {string} URL
+  # @param {string} FILE_SHA256 Content hash to verify downloaded file against.
```
kirkrodrigues, line 202:
```suggestion
  # @param {string} FILE_SHA256 Content hash to verify the downloaded file against.
```

taskfiles/utils.yml
```diff
@@ -141,3 +148,158 @@ tasks:
...
+  # @param {string} OUTPUT_DIR Directory in which to extract the tarball file.
+  # @param {string} URL
+  # @param {string} FILE_SHA256 Content hash to verify downloaded tar file against.
+  # @param {string={{.OUTPUT_DIR}}.md5} [CHECKSUM_FILE] File path to store the checksum of
+  # downloaded tar file.
+  # @param {string[]=[]} [EXCLUDE_PATTERNS] Path wildcard patterns that should not be extracted.
+  # @param {string[]=[]} [DATA_PATTERNS] Path wildcard patterns to extract.
+  # @param {int=1} [NUM_COMPONENTS_TO_STRIP] Number of leading path components to strip from the
+  # extracted files.
+  # @param {string={{.OUTPUT_DIR}}.tar.gz} [TAR_FILE] Path where the tarball file should be stored.
```
kirkrodrigues, line 254:
```suggestion
  # Runs curl to download the tar file from the given URL and extracts its contents.
  #
  # @param {string} OUTPUT_DIR Directory in which to extract the tar file.
  # @param {string} URL
  # @param {string} FILE_SHA256 Content hash to verify the downloaded tar file against.
  # @param {string={{.OUTPUT_DIR}}.md5} [CHECKSUM_FILE] File path to store the checksum of
  # downloaded tar file.
  # @param {string[]=[]} [EXCLUDE_PATTERNS] Path wildcard patterns that should not be extracted.
  # @param {string[]=[]} [DATA_PATTERNS] Path wildcard patterns to extract.
  # @param {int=1} [NUM_COMPONENTS_TO_STRIP] Number of leading path components to strip from the
  # extracted files.
  # @param {string={{.OUTPUT_DIR}}.tar.gz} [TAR_FILE] Path where the tar file should be stored.
```

Upon re-reading, I realized `tar file` is better.

taskfiles/utils.yml
```diff
@@ -141,3 +148,158 @@ tasks:
...
+        {{default (printf "%s.md5" .OUTPUT_DIR) .CHECKSUM_FILE}}
+      EXCLUDE_PATTERNS:
+        ref: "default (list) .EXCLUDE_PATTERNS"
+      DATA_PATTERNS:
+        ref: "default (list) .DATA_PATTERNS"
+      NUM_COMPONENTS_TO_STRIP: "{{default 1 .NUM_COMPONENTS_TO_STRIP}}"
+      TAR_FILE: >-
+        {{default (printf "%s.tar.gz" .OUTPUT_DIR) .TAR_FILE}}
+    requires:
+      vars: ["OUTPUT_DIR", "URL", "FILE_SHA256"]
```
kirkrodrigues, line 269: Alphabetize

taskfiles/utils.yml
```diff
@@ -141,3 +148,158 @@ tasks:
...
+    sources: ["{{.TASKFILE}}"]
+    generates: ["{{.CHECKSUM_FILE}}", "{{.TAR_FILE}}"]
+    deps:
+      - task: "curl"
+        vars:
+          URL: "{{.URL}}"
+          FILE_SHA256: "{{.FILE_SHA256}}"
+          OUTPUT_FILE: "{{.TAR_FILE}}"
```
kirkrodrigues, line 277: Alphabetize

kirkrodrigues (changes requested):

taskfiles/utils.yml
davidlion: Good catch. It works for extract, but not create.

taskfiles/utils.yml
```diff
@@ -1,64 +1,77 @@
...
+set: ["u", "pipefail"]
+shopt: ["globstar"]
+
 tasks:
   # ===
   # CHECKSUM UTILS
   # ===
 
-  # @param {string} DATA_DIR The directory to compute the checksum for.
+  # @param {string[]} DATA_PATTERNS Path wildcard patterns to compute the checksum for.
```
kirkrodrigues, line 11: What do you think about `INCLUDE_PATTERNS` to match `EXCLUDE_PATTERNS`?

taskfiles/utils.yml
```diff
@@ -1,64 +1,77 @@
...
       - >-
         tar cf -
-        --directory "{{.DATA_DIR}}"
         --group 0
         --mtime "UTC 1970-01-01"
         --numeric-owner
         --owner 0
         --sort name
-        {{- range .EXCLUDE_PATHS}}
+        --no-anchored
```
kirkrodrigues, line 30: Can we add a comment about why this is necessary?

taskfiles/utils.yml
```diff
@@ -1,64 +1,77 @@
...
+        --wildcards
+        {{- range .EXCLUDE_PATTERNS}}
         --exclude="{{.}}"
```
kirkrodrigues, line 33: Can we add a comment about why the exclude pattern needs quotes but the include pattern doesn't?

taskfiles/utils.yml
```diff
@@ -1,64 +1,77 @@
...
       - defer: "rm -f '{{.TMP_CHECKSUM_FILE}}'"
-      # Check that the directory exists and the checksum matches; otherwise delete the checksum file
-      - >-
-        (
-        test -d "{{.DATA_DIR}}"
-        && diff -q '{{.TMP_CHECKSUM_FILE}}' '{{.CHECKSUM_FILE}}' 2> /dev/null
-        ) || rm -f '{{.CHECKSUM_FILE}}'
+      # Check that all paths exist and the checksum matches; otherwise delete the checksum file.
+      - |-
+        ( {{- range .DATA_PATTERNS}}
```
kirkrodrigues, line 68:
```suggestion
        (
          {{- range .DATA_PATTERNS}}
```

taskfiles/utils.yml
```diff
@@ -141,3 +154,158 @@ tasks:
...
+    cmds:
+      - >-
+        cmake
+        -S "{{.SOURCE_DIR}}"
+        -B "{{.BUILD_DIR}}"
+        {{.CMAKE_ARGS}}
+      - >-
+        cmake
+        --build "{{.BUILD_DIR}}"
+        --parallel
```
kirkrodrigues, line 184: Can we add an optional arg to control the number of jobs (useful for memory constrained environments)?

taskfiles/utils.yml
```diff
@@ -141,3 +154,158 @@ tasks:
...
+    generates: ["{{.OUTPUT_FILE}}"]
+    status:
+      - >-
+        diff
+        <(echo "{{.FILE_SHA256}}")
+        <(openssl dgst -sha256 "{{.OUTPUT_FILE}}"
+        | awk '{print $2}')
+    cmds:
+      - |-
+        mkdir -p "{{dir .OUTPUT_FILE}}"
```
kirkrodrigues, line 226: Not mandatory, but to follow our conventions, we should use `--parents`

taskfiles/utils.yml
```diff
@@ -141,3 +154,158 @@ tasks:
...
+          echo "Failed to download after $max_attempts attempts."
+          exit 1
+        fi
+
+  # Runs curl to download the tar file from the given URL and extracts its contents.
+  #
+  # @param {string} OUTPUT_DIR Directory in which to extract the tar file.
+  # @param {string} URL
+  # @param {string} FILE_SHA256 Content hash to verify the downloaded tar file against.
+  # @param {string={{.OUTPUT_DIR}}.md5} [CHECKSUM_FILE] File path to store the checksum of
```
kirkrodrigues, line 254:
```suggestion
  # @param {string={{.OUTPUT_DIR}}.md5} [CHECKSUM_FILE] File path to store the checksum of the
```

taskfiles/utils.yml
```diff
@@ -141,3 +154,158 @@ tasks:
...
+          OUTPUT_FILE: "{{.TAR_FILE}}"
+          URL: "{{.URL}}"
+      - task: "validate-checksum"
+        vars:
+          CHECKSUM_FILE: "{{.CHECKSUM_FILE}}"
+          DATA_PATTERNS: ["{{.OUTPUT_DIR}}"]
+    cmds:
+      - |-
+        rm -rf "{{.OUTPUT_DIR}}"
+        mkdir -p "{{.OUTPUT_DIR}}"
```
kirkrodrigues, line 291: Not mandatory, but could use long form flags.

taskfiles/utils.yml
```diff
@@ -141,3 +154,158 @@ tasks:
...
+        ref: "default (list) .DATA_PATTERNS"
+      EXCLUDE_PATTERNS:
+        ref: "default (list) .EXCLUDE_PATTERNS"
+      NUM_COMPONENTS_TO_STRIP: "{{default 1 .NUM_COMPONENTS_TO_STRIP}}"
+      TAR_FILE: >-
+        {{default (printf "%s.tar.gz" .OUTPUT_DIR) .TAR_FILE}}
+    requires:
+      vars: ["FILE_SHA256", "OUTPUT_DIR", "URL"]
+    sources: ["{{.TASKFILE}}"]
+    generates: ["{{.CHECKSUM_FILE}}", "{{.TAR_FILE}}"]
```
kirkrodrigues, line 277: I need to double-check if these will cause issues.

kirkrodrigues (changes requested):

taskfiles/utils.yml
kirkrodrigues: * I guess you want `.TASKFILE` to be a source so that if the way of extracting the file changes, it will be extracted again, right?
* I think `.TARFILE` should be a source (rather than a generated file) since it's the generated file of `curl`.

taskfiles/utils.yml
```diff
@@ -1,64 +1,87 @@
...
       - defer: "rm -f '{{.TMP_CHECKSUM_FILE}}'"
-      # Check that the directory exists and the checksum matches; otherwise delete the checksum file
-      - >-
+      # Check that all paths exist and the checksum matches; otherwise delete the checksum file.
+      - |-
         (
-        test -d "{{.DATA_DIR}}"
-        && diff -q '{{.TMP_CHECKSUM_FILE}}' '{{.CHECKSUM_FILE}}' 2> /dev/null
-        ) || rm -f '{{.CHECKSUM_FILE}}'
+          {{- range .DATA_PATTERNS}}
```
kirkrodrigues, line 78:
```suggestion
          {{- range .INCLUDE_PATTERNS}}
```

taskfiles/utils.yml
```diff
@@ -1,64 +1,87 @@
...
+        {{.}}
+        {{- end}}
+        2> /dev/null
         | md5sum > {{.OUTPUT_FILE}}
     # Ignore errors so that dependent tasks don't fail
     ignore_error: true
 
-  # @param {string} DATA_DIR The directory to validate the checksum for.
+  # @param {string[]} INCLUDE_PATTERNS Path wildcard patterns to validate the checksum for.
   # @param {string} OUTPUT_FILE
```
kirkrodrigues, line 53:
```suggestion
  # @param {string} CHECKSUM_FILE
  # @param {string[]} INCLUDE_PATTERNS Path wildcard patterns to validate the checksum for.
```

Note the rename from OUTPUT_FILE to CHECKSUM_FILE.

taskfiles/utils.yml
```diff
@@ -141,3 +164,164 @@ tasks:
...
+    vars:
+      CHECKSUM_FILE: >-
+        {{default (printf "%s.md5" .OUTPUT_DIR) .CHECKSUM_FILE}}
+      INCLUDE_PATTERNS:
+        ref: "default (list) .INCLUDE_PATTERNS"
+      EXCLUDE_PATTERNS:
+        ref: "default (list) .EXCLUDE_PATTERNS"
+      NUM_COMPONENTS_TO_STRIP: "{{default 1 .NUM_COMPONENTS_TO_STRIP}}"
+      TAR_FILE: >-
+        {{default (printf "%s.tar.gz" .OUTPUT_DIR) .TAR_FILE}}
```
kirkrodrigues, line 287:
```suggestion
      # Paths
      CHECKSUM_FILE: >-
        {{default (printf "%s.md5" .OUTPUT_DIR) .CHECKSUM_FILE}}
      TAR_FILE: >-
        {{default (printf "%s.tar.gz" .OUTPUT_DIR) .TAR_FILE}}
      
      # Path patterns
      EXCLUDE_PATTERNS:
        ref: "default (list) .EXCLUDE_PATTERNS"
      INCLUDE_PATTERNS:
        ref: "default (list) .INCLUDE_PATTERNS"
      
      # Misc
      NUM_COMPONENTS_TO_STRIP: "{{default 1 .NUM_COMPONENTS_TO_STRIP}}"
```

Our guidelines ask that [map entries be ordered alphabetically](https://docs.yscope.com/dev-guide/contrib-guides-taskfiles.html#ordering-of-list-map-entries), within each group, so I think we can break this into groups.

taskfiles/utils.yml
```diff
@@ -1,64 +1,87 @@
...
-      vars: ["DATA_DIR", "OUTPUT_FILE"]
+      vars: ["INCLUDE_PATTERNS", "OUTPUT_FILE"]
     cmds:
+      # We explicitly set --no-anchored and --wildcards to ensure the behaviour for both exclusion
+      # and inclusion matches, as they are only the default for exclusion.
+      # We can quote the exclude patterns for tar because the patterns themselves are passed to tar
+      # as arguments and evaluated by tar during execution. However, for tar --create the input
+      # patterns cannot be quoted as they are evaluated by the shell and the results are passed to
+      # tar as arguments. If the input patterns are passed to tar with quotes the pattern will not
+      # be evaluated and will be treated literally.
```
kirkrodrigues, line 29:
```suggestion
      # We explicitly set `--no-anchored` and `--wildcards` to make the inclusion behaviour match
      # the default exclusion behaviour.
      #
      # We can quote the exclude patterns for `tar` because the patterns themselves are passed to
      # `tar` as arguments and evaluated by `tar` during execution. However, for `tar --create`, the
      # input patterns cannot be quoted since they're evaluated by the shell and the results are
      # passed to `tar` as arguments. If the input patterns are passed to `tar` with quotes, the
      # pattern won't be evaluated and will instead be treated literally.
```

taskfiles/utils.yml
```diff
@@ -1,64 +1,87 @@
...
     # Ignore errors so that dependent tasks don't fail
     ignore_error: true
 
-  # @param {string} DATA_DIR The directory to validate the checksum for.
+  # @param {string[]} INCLUDE_PATTERNS Path wildcard patterns to validate the checksum for.
   # @param {string} OUTPUT_FILE
-  # @param {[]string} [EXCLUDE_PATHS] A list of paths, relative to `DATA_DIR`, to exclude from the
-  # checksum.
+  # @param {string[]} [EXCLUDE_PATTERNS] Path wildcard patterns, relative to any `INCLUDE_PATTERNS`, to
+  # exclude from the checksum.
```
kirkrodrigues, line 55:
```suggestion
  # @param {string[]} [EXCLUDE_PATTERNS] Path wildcard patterns, relative to any `INCLUDE_PATTERNS`,
  # to exclude from the checksum.
```

taskfiles/utils.yml
```diff
@@ -141,3 +164,164 @@ tasks:
...
+  # ===
+  # CMAKE UTILS
+  # ===
+
+  # Runs CMake's configure and build steps for the given source and build directories.
+  #
+  # @param {string} BUILD_DIR CMake build directory to create.
+  # @param {string} SOURCE_DIR Project source directory containing the CMakeLists.txt file.
+  # @param {int} [JOBS] The maximum number of concurrent processes to use when building. If
+  # omitted the native build tool's default number is used. See man cmake.
```
kirkrodrigues, line 177:
```suggestion
  # omitted, the native build tool's default number is used. See `man cmake`.
```

taskfiles/utils.yml
```diff
@@ -141,3 +164,164 @@ tasks:
...
+  # @param {string} [CONF_ARGS] Any additional arguments to pass to CMake's configure step.
+  cmake-config-and-build:
+    internal: true
+    label: "{{.TASK}}-{{.SOURCE_DIR}}-{{.BUILD_DIR}}"
+    vars:
+      JOBS: >-
+        {{default "" .JOBS}}
+      CONF_ARGS: >-
+        {{default "" .CONF_ARGS}}
```
kirkrodrigues, line 186:
```suggestion
      CONF_ARGS: >-
        {{default "" .CONF_ARGS}}
      JOBS: >-
        {{default "" .JOBS}}
```

taskfiles/utils.yml
```diff
@@ -141,3 +164,164 @@ tasks:
...
+      - task: "validate-checksum"
+        vars:
+          CHECKSUM_FILE: "{{.CHECKSUM_FILE}}"
+          INCLUDE_PATTERNS: ["{{.OUTPUT_DIR}}"]
+    cmds:
+      - |-
+        rm --force --recursive "{{.OUTPUT_DIR}}"
+        mkdir --parents  "{{.OUTPUT_DIR}}"
+      # Unlike tar --create, in --extract both the exclude and include patterns can be quoted as
+      # they are both evaluted by tar.
```
kirkrodrigues, line 307:
```suggestion
      # Unlike for `tar --create`, in `--extract` both the exclude and include patterns can be
      # quoted since they're both evaluated by `tar`.
```

kirkrodrigues (changes requested):
Mainly style and one error.

For the PR title, how about:

```
feat(taskfiles): Add cmake and remote utils; Update checksum data path to an array.
```

kirkrodrigues (approved):

PR #17, opened 2025-01-18
feat(cli): Add cli script for running clang-tidy in parallel.

LinZhihao-723:
# Description

This PR introduces a cli sub-project for running clang-tidy in parallel. For now, it is almost like a simplified script of the llvm version [here](https://github.com/llvm/llvm-project/blob/main/clang-tools-extra/clang-tidy/tool/run-clang-tidy.py#L499-L534).

# Validation performed

I have tested this cli by using it to check clp-ffi-py. Here's how it works:
```shell
python3 -m venv venv
. ./venv/bin/activate
cd $PROJECT_DIR/yscope-dev-utils/cli/clang-tidy-utils
pip install .
cd $PROJECT_DIR/clp-ffi-py
task lint:cpp-format-fix # This is to ensure clang-tidy config has been set
rm -rf build && mkdir build && cd build && cmake .. && cd ..
yscope-clang-tidy-utils -j 10 ./src/clp_ffi_py/ir/native/* -- -p build/compile_commands.json  --line-filter="[{\"name\":\"cpp11_zone.hpp\",\"lines\":[[197,197]]}]"
```
Here, we parallel 10 clang-tidy tasks with extra clang-tidy arguments `-p build/compile_commands.json  --line-filter="[{\"name\":\"cpp11_zone.hpp\",\"lines\":[[197,197]]}]"` to match what we specify in the linting task [here](https://github.com/y-scope/clp-ffi-py/blob/2595c70b93b34d7326039a97e723d6e4fbc8d893/lint-tasks.yml#L107). You can use `echo $?` to check the return value is 0. If we remove the lint filter in the arguments above, clang-tidy will raise a warning on msgpack's source code and thus the return value of `echo $?` will be 1 in this case.

### DIFF
--- a/cli/clang-tidy-utils/pyproject.toml
+++ b/cli/clang-tidy-utils/pyproject.toml
@@ -1,0 +1,50 @@
+[build-system]
+requires = ["setuptools >= 61.0", "setuptools_scm"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "yscope-clang-tidy-utils"
+version = "0.0.1"
+authors = [
+    { name="zhihao lin", email="lin.zhihao@yscope.com" },
+]
+requires-python = ">=3.10"
+dependencies = [
+    "clang-tidy >= 19.1.0",
+]
+
+[project.optional-dependencies]
+dev = [
+    "black >= 24.10.0",
+    "docformatter >= 1.7.5",
+    "mypy >= 1.14.1",
+    "ruff >= 0.9.2",
+]
+
+[project.scripts]
+yscope-clang-tidy-utils = "yscope_clang_tidy_utils.cli:main"
+
+[tool.black]
+line-length = 100
+target-version = ["py311"]
+color = true
+preview = true
+
+[tool.docformatter]
+make-summary-multi-line = true
+pre-summary-newline = true
+recursive = true
+wrap-summaries = 100
+wrap-descriptions = 100
+
+[tool.mypy]
+explicit_package_bases = true
+strict = true
+pretty = true
+
+[tool.ruff]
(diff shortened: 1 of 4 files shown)


cli/clang-tidy-utils/pyproject.toml
```diff
@@ -0,0 +1,45 @@
+[project]
+name = "yscope-clang-tidy-utils"
```
LinZhihao-723, line 2: We may drop `yscope-` prefix

LinZhihao-723 (commented):
We can discuss how to structure the cli scripts/tools. The current layout might not be the best practice.

cli/clang-tidy-utils/pyproject.toml
kirkrodrigues: I prefer to drop the prefix for now since it's likely this will be an internal tool for the foreseeable future.

kirkrodrigues (commented):
Not a complete review but a few quick comments:

- Can we move the tool into `linters` rather than `cli`?
- Can we prefix internal functions with `_`?
- There are some spelling/grammar mistakes in the README. Can you go over it once more?

cli/clang-tidy-utils/pyproject.toml
LinZhihao-723: Removed.

LinZhihao-723:
> Not a complete review but a few quick comments:
> 
> * Can we move the tool into `linters` rather than `cli`?
> * Can we prefix internal functions with `_`?
> * There are some spelling/grammar mistakes in the README. Can you go over it once more?

Fixed.

PR #20, opened 2025-02-12
chore(taskfiles)!: Split up cmake-config-and-build; Add cmake-clean.

davidlion:
# Description

This PR breaks up `cmake-config-and-build` into `cmake-generate` and `cmake-build`. Additionally, it adds `cmake-clean` and improves the docstrings for each task.

The notable decisions in our design are:
* We do not use `sources` or `generates` and defer decision making to `cmake` to ensure expected and reliable behaviour.
* We leave it up to the user/caller to ensure the proper dependencies between tasks are met (e.g. calling generate before build).
    * Adding dependencies causes a task to require all the arguments of its dependent tasks to be run correctly (i.e. install becomes a superset of build and generate, build becomes a superset of generate). This can lead to unexpected behaviour as a change in any argument could lead to re-generating and re-building the project with different arguments.

# Checklist

  breaking change.

# Validation performed

Tested with various tasks in `clp`.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `cmake-clean` command to enhance cleaning operations.
	- Expanded command options with `EXTRA_ARGS` and `TARGETS` for more flexible build configurations.

- **Chores**
	- Updated task definitions and clarified dependencies for improved workflow understanding.

### DIFF
--- a/taskfiles/utils-cmake.yaml
+++ b/taskfiles/utils-cmake.yaml
@@ -4,21 +4,59 @@ set: ["u", "pipefail"]
 shopt: ["globstar"]
 
 tasks:
-  # Runs CMake's configure and build steps for the given source and build directories.
+  # Runs the CMake build step for the given build directory. The caller must have previously called
+  # `cmake-generate` on `BUILD_DIR` for this task to succeed. We purposely omit `sources` and
+  # `generates` as we defer to `cmake` to decide whether it should perform any actions.
   #
-  # @param {string} BUILD_DIR CMake build directory to create.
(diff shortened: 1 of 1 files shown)


kirkrodrigues (changes requested):
Minor refactoring.

Bill-hbrhbr:
Changes lgtm. My only concern is that, although not time-consuming, CMake build and install will dump a lot of useless info when source files have not been changed.
Maybe on a higher level we should still use `sources` and/or `generates` to guard whether the `generate->build->install` flow should be run or not.

kirkrodrigues (approved):

PR #24, opened 2025-02-14
fix: Force `utils:curl` to run when input urls or commit hash for the library target to be downloaded changes.

Bill-hbrhbr:
# Description

# Checklist

* [X] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [X] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [X] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

* [X] Tested extensively in [`ystdlib-cpp`](https://github.com/y-scope/ystdlib-cpp/pull/25) by stimulating the adjusting of input arguments to `utils:curl`.

### DIFF
--- a/taskfiles/utils-remote.yaml
+++ b/taskfiles/utils-remote.yaml
@@ -14,10 +14,15 @@ tasks:
     label: "{{.TASK}}-{{.OUTPUT_FILE}}"
     vars:
       OUTPUT_FILE: "{{default (base .URL) .OUTPUT_FILE}}"
+      URL_FILE: "{{.OUTPUT_FILE | replace \".\" \"#\"}}.url"
     requires:
       vars: ["FILE_SHA256", "URL"]
-    generates: ["{{.OUTPUT_FILE}}"]
+    generates: ["{{.OUTPUT_FILE}}, {{.URL_FILE}}"]
     status:
+      - >-
+        diff
+        <(echo "{{.URL}}")
+        <(cat "{{.URL_FILE}}")
       - >-
         diff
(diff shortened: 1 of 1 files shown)


davidlion (commented):
Any reason we can't just add the SHA to the task `label` rather than storing a separate file?

Also, is it possible to just add some output in `status`?

Bill-hbrhbr:
> Any reason we can't just add the SHA to the task `label` rather than storing a separate file?
> 
> Also, is it possible to just add some output in `status`?

I think the `label` works well with `sources`, but `status` operates independently, disregarding the label of the `task`. I can showcase this behavior with a dummy example.
Also what kinds of output are you thinking in `status`? Do you mean extra logging statements (e.g. on the checks we are performing)?

davidlion:
Closing this issue for the following reasons:

> Task can't force re-runs by detecting input argument changes. Without this PR, utils:curl will stay ' up-to-date' if we change only the url of the tar source but not its FILE_SHA256.

Afaik, there shouldn't be a need to re-download the file from a new/different URL if the existing file (`OUTPUT_FILE`) matches the specified `FILE_SHA256` (i.e. `status` succeeds) as any newly downloaded file would be identical to what we already have.
For testing purposes we can use the `-f/--force` flag for `task` to ensure tasks are re-run.

> Also in such situations where the tar source downloaded doesn't match the given FILE_SHA256, no error message is outputted. The task simply re-runs over and over again without users knowing that they have supplied a wrong FILE_SHA256.

This fix has been separated and moved into #70.

PR #25, opened 2025-02-18
feat(taskfiles)!: Add filtering support to `clang-tidy` and `clang-format` tasks; Add support for running `clang-tidy` on multiple files concurrently.

davidlion:
# Description

This PR adds/updates 3 "high-level" tasks to be used by other projects:
1. `clang-format`: updated to have multiple pattern parameters to `find` allowing for precise filtering on files to run on
2. `clang-tidy-find`:  similar to `clang-format` uses `find` to filter files to run on
3. `clang-tidy-diff`: uses `git diff` to run `clang-tidy` on any files differing from a given git ref

The remaining added tasks are used by 2 and 3 so that clang-tidy:
* is run concurrently (through multiple tasks) on each file (parallelism can be controlled through task)
* outputs log files for each file and displays completed logs on failure

# Checklist

  breaking change.

# Validation performed

Tested on `clp` locally with new tasks to be PR'd later.

## Summary by CodeRabbit

- **New Features**
  - Added new linting tasks: `clang-tidy-diff`, `clang-tidy-find`, `clang-tidy-output`, and `clang-tidy-parallel`.

- **Enhancements**
  - Improved `clang-format` and `clang-tidy` tasks with better filtering options and detailed parameter descriptions.
  - Enhanced output management for linting tasks, providing clearer indicators of task success or failure.


taskfiles/utils-cpp-lint.yaml
```diff
@@ -6,34 +6,284 @@ shopt: ["globstar"]
...
-  # @param {string} VENV_DIR Python virtual environment where clang-format is installed.
+  # @param {string[]} ROOT_PATHS The paths on which to start `find`. When combined with
+  # EXCLUDE_PATTERNS and INCLUDE_PATTERNS `find` will output a list of files used as the input to
+  # `clang-format`.
+  # @param {string} VENV_DIR Python virtual environment where `clang-format` is installed.
+  # @param {string[]} [EXCLUDE_PATTERNS] `-path` wildcard patterns to prune in `find`.
+  # @param {string[]} [EXTENSION_PATTERNS=["*.cpp", "*.h", "*.hpp"]] `-name` wildcard patterns to
+  # match in `find`.
+  # @param {string} [FLAGS] Any flags to pass to `clang-format`.
+  # @param {string[]} [INCLUDE_PATTERNS] `-path` wildcard patterns to match in `find`.
```
kirkrodrigues, line 17: * Can we remove `find` implementation details?
* Can we reorder the arguments to be more logical? Perhaps:
  * `VENV_DIR`
  * `ROOT_PATHS`
  * `INCLUDE_PATTERNS`
  * `EXCLUDE_PATTERNS`
  * `EXTENSION_PATERNS`
  * `FLAGS`

taskfiles/utils-cpp-lint.yaml
```diff
@@ -6,34 +6,284 @@ shopt: ["globstar"]
...
-  # @param {string[]} SRC_PATHS The paths on which to run clang-format.
-  # @param {string} VENV_DIR Python virtual environment where clang-format is installed.
+  # @param {string[]} ROOT_PATHS The paths on which to start `find`. When combined with
+  # EXCLUDE_PATTERNS and INCLUDE_PATTERNS `find` will output a list of files used as the input to
+  # `clang-format`.
+  # @param {string} VENV_DIR Python virtual environment where `clang-format` is installed.
+  # @param {string[]} [EXCLUDE_PATTERNS] `-path` wildcard patterns to prune in `find`.
+  # @param {string[]} [EXTENSION_PATTERNS=["*.cpp", "*.h", "*.hpp"]] `-name` wildcard patterns to
+  # match in `find`.
+  # @param {string} [FLAGS] Any flags to pass to `clang-format`.
```
kirkrodrigues, line 16:
```suggestion
  # @param {string[]} [FLAGS] Any flags to pass to `clang-format`.
```

taskfiles/utils-cpp-lint.yaml
```diff
@@ -6,34 +6,284 @@ shopt: ["globstar"]
...
+    vars:
+      EXCLUDE_PATTERNS:
+        ref: "default (list) .EXCLUDE_PATTERNS"
+      EXTENSION_PATTERNS:
+        ref: >-
+          default (list "*.cpp" "*.h" "*.hpp") .EXTENSION_PATTERNS
+      FLAGS:
+        ref: "default (list) .FLAGS"
+      INCLUDE_PATTERNS:
+        ref: "default (list) .INCLUDE_PATTERNS"
```
kirkrodrigues, line 29: How about grouping the path-related patterns together under a comment: `# Path patterns`?

taskfiles/utils-cpp-lint.yaml
```diff
@@ -6,34 +6,284 @@ shopt: ["globstar"]
...
     requires:
-      vars: ["FLAGS", "SRC_PATHS", "VENV_DIR"]
+      vars: ["ROOT_PATHS", "VENV_DIR"]
     cmd: |-
```
kirkrodrigues, line 32: To be more readable, can we use `>-` and then get rid of the escaped line breaks?

Something like:
```yaml
. "{{.VENV_DIR}}/bin/activate
&& find
   {{- range .ROOT_PATHS}} ...
```

taskfiles/utils-cpp-lint.yaml
```diff
@@ -6,34 +6,284 @@ shopt: ["globstar"]
...
       . "{{.VENV_DIR}}/bin/activate"
-      find {{- range .SRC_PATHS}} "{{.}}" {{- end}} \
+      find \
+        {{- range .ROOT_PATHS}}
+        {{.}} \
+        {{- end}}
```
kirkrodrigues, line 37: Could combine this into a single line?

taskfiles/utils-cpp-lint.yaml
```diff
@@ -6,34 +6,284 @@ shopt: ["globstar"]
...
+        {{- if .EXCLUDE_PATTERNS}}
+          \( \ {{- range $i, $pattern := .EXCLUDE_PATTERNS}}
+            {{if $i}}-or {{end -}}
+             -path "{{$pattern}}" \
+            {{- end}}
```
kirkrodrigues, line 42: Shall we continue indenting if-end blocks and perhaps start indenting range-end blocks?

kirkrodrigues (changes requested):
Review of the `clang-format` task. Can you see if you agree with my suggestions and if so, then apply them to the other tasks before a re-review?

davidlion:
Accidentally deleted the branch... sorry.

taskfiles/utils-cpp-lint.yaml
davidlion: Unfortunately, indentation causes newlines to not fold in yaml, so there is no way to indent the text without causing shell syntax errors.

taskfiles/utils-cpp-lint.yaml
davidlion: Went with full indentation.

taskfiles/utils/cpp-lint.yaml
```diff
@@ -4,36 +4,324 @@ set: ["u", "pipefail"]
...
+        -print0
+      | xargs -0 clang-format \
+        {{- range .FLAGS}}
+          {{.}} \
+        {{- end}}
+        -Werror
 
-  # Runs clang-tidy on C++ files at the given paths.
+  # Run `clang-tidy` on `SOURCE_PATH` and write the output to `OUTPUT_PATH`. The output is appended
+  # with "Passed" or "Failed" depending on the result.
```
kirkrodrigues, line 89:
```suggestion
  # Runs `clang-tidy` on `SOURCE_PATH` and writes the output to `OUTPUT_PATH`. The strings "Passed"
  # or "Failed" are appended to the output file, depending on the result.
```

taskfiles/utils/cpp-lint.yaml
```diff
@@ -4,36 +4,324 @@ set: ["u", "pipefail"]
...
+  clang-tidy-diff:
+    internal: true
+    label: "{{.TASK}}:{{.OUTPUT_DIR}}"
+    vars:
+      GIT_REF: >-
+        {{default "origin/main" .GIT_REF}}
+      DIFF_SOURCE_PATHS_:
+        sh: >-
+          git diff --name-only --ignore-submodules "{{.GIT_REF}}"
+          | grep ".*\.\(cpp\|h\|hpp\)"
```
kirkrodrigues, line 139:
```suggestion
          git diff --name-only --ignore-submodules "{{.GIT_REF}}" **/*.{cpp,h,hpp}
```

That should work, right?

taskfiles/utils/cpp-lint.yaml
```diff
@@ -4,36 +4,324 @@ set: ["u", "pipefail"]
...
+          OUTPUT_DIR: "{{.OUTPUT_DIR}}"
+          SOURCE_PATHS:
+            ref: ".SOURCE_PATHS"
+          VENV_DIR: "{{.VENV_DIR}}"
+
+  # Run `clang-tidy-parallel` and print completed output logs on failure. Output logs that were not
+  # fully completed or do not exist (`clang-tidy` did not get a chance to run for the file) are
+  # skipped.
+  #
+  # @param {string} SOURCE_PATHS List of source file paths.
```
kirkrodrigues, line 263:
```suggestion
  # @param {string[]} SOURCE_PATHS List of source file paths.
```

kirkrodrigues, line 263: How about `SOURCE_FILE_PATHS`? Then I think you can drop the description.

taskfiles/utils/cpp-lint.yaml
```diff
@@ -4,36 +4,324 @@ set: ["u", "pipefail"]
...
+  # @param {string} VENV_DIR Python virtual environment where `clang-tidy` is installed.
+  # @param {string[]} [FLAGS] Any flags to pass to `clang-tidy`.
+  # @param {string} [OUTPUT_DIR="{{.ROOT_DIR}}/lint-clang-tidy"] Directory to store the `clang-tidy`
+  # output for each file.
```
kirkrodrigues, line 267: I think you can drop the description?

taskfiles/utils/cpp-lint.yaml
```diff
@@ -4,36 +4,324 @@ set: ["u", "pipefail"]
...
+            {{- $output_dir := .OUTPUT_DIR}}
+            {{- range .SOURCE_PATHS}}
+              {{- $output_path := printf "%s/%s.log" $output_dir (replace "/" "-" .)}}
+              if [[ -f "{{$output_path}}" ]] \
+                  && [[ "Failed" == "$(tail -n 1 {{$output_path}})" ]]; then
+                tail -v -n +1 "{{$output_path}}"
+              fi
+            {{- end}}
+          {{- end}}
+      - mkdir -p "{{.OUTPUT_DIR}}"
```
kirkrodrigues, line 289:
```suggestion
      - "mkdir -p '{{.OUTPUT_DIR}}'"
```

taskfiles/utils/cpp-lint.yaml
```diff
@@ -4,36 +4,324 @@ set: ["u", "pipefail"]
...
+            ref: ".FLAGS"
+          OUTPUT_DIR: "{{.OUTPUT_DIR}}"
+          SOURCE_PATHS:
+            ref: ".SOURCE_PATHS"
+          VENV_DIR: "{{.VENV_DIR}}"
+
+  # Run `clang-tidy` in parallel on `SOURCE_PATHS`, storing the output for each file in
+  # `OUTPUT_DIR`. Each file is named after the source file path with '/' replaced with '-' and
+  # suffixed with '.log'. The task will fail on the first file to fail `clang-tidy`, which will
+  # result in some files having incomplete output.
```
kirkrodrigues, line 302:
```suggestion
  # Runs `clang-tidy` in parallel on `SOURCE_PATHS`, storing the output for each file in
  # `OUTPUT_DIR`. Each output file is named after the source file path except '/' is replaced with
  # '-' and the file is suffixed with '.log'. The task will fail on the first file to fail
  # `clang-tidy`, which may result in some files having incomplete output.
```

taskfiles/utils/cpp-lint.yaml
```diff
@@ -4,36 +4,324 @@ set: ["u", "pipefail"]
...
+  #
+  # @param {string} SOURCE_PATHS List of source file paths.
```
kirkrodrigues, line 304:
```suggestion
  # @param {string[]} SOURCE_PATHS List of source file paths.
```

kirkrodrigues, line 304: How about `SOURCE_FILE_PATHS`? Then I think you can drop the description.

taskfiles/utils/cpp-lint.yaml
```diff
@@ -4,36 +4,324 @@ set: ["u", "pipefail"]
...
+  # @param {string} VENV_DIR Python virtual environment where `clang-tidy` is installed.
+  # @param {string[]} [FLAGS] Any flags to pass to `clang-tidy`.
+  # @param {string} [OUTPUT_DIR="{{.ROOT_DIR}}/lint-clang-tidy"] Directory to store the `clang-tidy`
+  # output for each file.
```
kirkrodrigues, line 308: I think we can drop this description.

taskfiles/utils/cpp-lint.yaml
```diff
@@ -4,36 +4,324 @@ set: ["u", "pipefail"]
...
+  #
+  # Pattern parameters
+  # @param {string[]} [EXCLUDE_PATTERNS] Wildcard patterns where any matching paths (relative to
+  # `ROOT_PATHS`) will be excluded. The search will not descend into matching directory paths. By
+  # default no paths match.
+  # @param {string[]} [EXTENSION_PATTERNS=["*.cpp", "*.h", "*.hpp"]] Wildcard patterns matched
+  # against the file name (not the full path) determining files to run on.
+  # @param {string[]} [INCLUDE_PATTERNS] Wildcard patterns where any matching file paths (relative
+  # to `ROOT_PATHS`) will be included. By default all paths match.
+  clang-tidy-find:
```
kirkrodrigues, line 181: Skipped reviewing since it seems similar to the clang-format task and will likely require similar refactoring.

taskfiles/utils/cpp-lint.yaml
```diff
@@ -4,36 +4,324 @@ set: ["u", "pipefail"]
...
+  # @param {string[]} [FLAGS] Any flags to pass to `clang-format`.
+  #
+  # Pattern parameters
+  # @param {string[]} [EXCLUDE_PATTERNS] Wildcard patterns where any matching paths (relative to
+  # `ROOT_PATHS`) will be excluded. The search will not descend into matching directory paths. By
+  # default no paths match.
+  # @param {string[]} [EXTENSION_PATTERNS=["*.cpp", "*.h", "*.hpp"]] Wildcard patterns matched
+  # against the file name (not the full path) determining files to run on.
+  # @param {string[]} [INCLUDE_PATTERNS] Wildcard patterns where any matching file paths (relative
+  # to `ROOT_PATHS`) will be included. By default all paths match.
```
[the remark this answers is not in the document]
davidlion: Good point. I'll just use `INCLUDE_FILENAME_PATTERNS` since it is more flexible and it requires less (no) extra string manipulation.

taskfiles/utils/cpp-lint.yaml
davidlion: Pretty sure you're right, but I'll double check.

taskfiles/utils/cpp-lint.yaml
```diff
@@ -4,36 +4,315 @@ set: ["u", "pipefail"]
...
+  # Runs `clang-tidy` in parallel on C++ source files that differ between HEAD and the target
+  # branch. On failure, `clang-tidy`'s output is printed. The task will fail on the first file which
+  # fails `clang-tidy`'s checks, which will result in some files having incomplete output. Users can
+  # either fix the known errors and run the task again or directly run `clang-tidy` on a file of
+  # interest.
+  #
+  # @param {string} VENV_DIR Python virtual environment where `clang-tidy` is installed.
+  # @param {string} [GIT_REF="origin/main"] A `git` reference that HEAD is diff'd with.
+  # @param {string[]} [FLAGS] Any flags to pass to `clang-tidy`.
+  # @param {string} [OUTPUT_DIR="{{.ROOT_DIR}}/lint-clang-tidy"]
```
LinZhihao-723, line 127: Quick question: it seems like this diff would only return files being changed directly. However, a change to a header file might affect other files implicitly. Shall we document this behavior somewhere?

taskfiles/utils/cpp-lint.yaml
```diff
@@ -4,36 +4,315 @@ set: ["u", "pipefail"]
...
+          {{- range .FLAGS}}
+            {{.}} \
+          {{- end}}
+          "{{.SOURCE_FILE_PATH}}" &> "{{.OUTPUT_FILE_PATH}}"
+
+  # Runs `clang-tidy` in parallel on C++ source files that differ between HEAD and the target
+  # branch. On failure, `clang-tidy`'s output is printed. The task will fail on the first file which
+  # fails `clang-tidy`'s checks, which will result in some files having incomplete output. Users can
+  # either fix the known errors and run the task again or directly run `clang-tidy` on a file of
+  # interest.
```
kirkrodrigues, line 122:
```suggestion
  # Runs `clang-tidy` in parallel on C++ source files that differ between HEAD and a given git
  # reference. On failure, `clang-tidy`'s output is printed. The task will fail on the first file
  # that fails `clang-tidy`'s checks, which will result in some files having incomplete output.
  # Users can either fix the known errors and run the task again or directly run `clang-tidy` on a
  # file of interest.
```

taskfiles/utils/cpp-lint.yaml
```diff
@@ -4,36 +4,315 @@ set: ["u", "pipefail"]
...
+  #
+  # @param {string} VENV_DIR Python virtual environment where `clang-tidy` is installed.
+  # @param {string} [GIT_REF="origin/main"] A `git` reference that HEAD is diff'd with.
```
kirkrodrigues, line 125:
```suggestion
  # @param {string} [GIT_REF="origin/main"] The `git` reference to diff against HEAD.
```

taskfiles/utils/cpp-lint.yaml
```diff
@@ -4,36 +4,315 @@ set: ["u", "pipefail"]
...
+        ref: "default (list) .EXCLUDE_PATTERNS"
+      INCLUDE_FILENAME_PATTERNS:
+        ref: >-
+          default (list "*.cpp" "*.h" "*.hpp") .INCLUDE_FILENAME_PATTERNS
+      INCLUDE_PATTERNS:
+        ref: "default (list) .INCLUDE_PATTERNS"
     requires:
-      vars: ["FLAGS", "SRC_PATHS", "VENV_DIR"]
+      vars: ["ROOT_PATHS", "VENV_DIR"]
     cmd: |-
```
kirkrodrigues, line 40:
```suggestion

    # NOTE: We don't quote `ROOT_PATHS` entries so that wildcards can be evaluated by the shell.
    cmd: |-
```

taskfiles/utils/cpp-lint.yaml
```diff
@@ -4,36 +4,315 @@ set: ["u", "pipefail"]
...
+        ref: "default (list) .EXCLUDE_PATTERNS"
+      INCLUDE_FILENAME_PATTERNS:
+        ref: >-
+          default (list "*.cpp" "*.h" "*.hpp") .INCLUDE_FILENAME_PATTERNS
+      INCLUDE_PATTERNS:
+        ref: "default (list) .INCLUDE_PATTERNS"
+
+      # Task variables
+      SOURCE_FILE_PATHS_:
+        # Do not quote `ROOT_PATHS` entries so that wildcards can be evaluated by the shell.
```
kirkrodrigues, line 196:
```suggestion
        # NOTE:
        # - We don't quote `ROOT_PATHS` entries so that wildcards can be evaluated by the shell.
        # - `find`'s `-print0` option adds a null after the last entry, so care must be taken
        #   when splitting the output using a null as a delimiter.
```

taskfiles/utils/cpp-lint.yaml
```diff
@@ -4,36 +4,315 @@ set: ["u", "pipefail"]
...
+                    -or \
+                  {{- end}}
+                  -name "{{$pattern}}" \
+                {{- end}}
+              \) \
+            {{- end}}
+            -print0
+      SOURCE_FILE_PATHS:
+        # `-print0` output ends in "\x00" causing `splitList` to generate an empty entry, that we
+        # remove using `compact`.
```
kirkrodrigues, line 238:
```suggestion
        # `SOURCE_FILE_PATHS_`'s last entry is an empty string which we remove using `compact`.
```

taskfiles/utils/cpp-lint.yaml
kirkrodrigues, line 122: Since `clang-tidy-find` is similar except that it works with path filtering, should we include the same details in its docstring (e.g., that clang-tidy is run in parallel)?

kirkrodrigues (changes requested):

taskfiles/utils/cpp-lint.yaml
davidlion: Ya you're right. I was thinking users could read the called tasks, but that seems like an unnecessary burden.

taskfiles/utils/cpp-lint.yaml
davidlion: As discussed offline:

This is a good point. Ideally we try to properly handle this, but it likely won't be simple. For now we will document the issue and let nightly complete runs of clang-tidy catch issues like this.

taskfiles/utils/cpp-lint.yaml
```diff
@@ -4,36 +4,331 @@ set: ["u", "pipefail"]
...
+
+  # Runs `clang-tidy` in parallel on C++ source files that differ between HEAD and a given git
+  # reference. On failure, `clang-tidy`'s output is printed.
+  #
+  # The task will fail on the first file that fails `clang-tidy`'s checks, which will result in some
+  # files having incomplete output. Users can either fix the known errors and run the task again or
+  # directly run `clang-tidy` on a file of interest.
+  #
+  # Note, that changes to a header file may cause failures in files that include it (directly or
+  # indirectly), making it possible for all modified to pass, but unmodified files to fail.
```
kirkrodrigues, line 128:
```suggestion
  # Note that changes to a header file may cause failures in files that include it (directly or
  # indirectly), making it possible for all modified files to pass, but unmodified files to fail.
```

taskfiles/utils/cpp-lint.yaml
```diff
@@ -4,36 +4,331 @@ set: ["u", "pipefail"]
...
+
+  # Runs `clang-tidy` in parallel on C++ source files that match the given filter criteria. On
+  # failure, `clang-tidy`'s output is printed.
+  #
+  # The task will fail on the first file that fails `clang-tidy`'s checks, which will result in some
+  # files having incomplete output. Users can either fix the known errors and run the task again or
+  # directly run `clang-tidy` on a file of interest.
+  #
+  # Note, that changes to a header file may cause failures in files that include it (directly or
+  # indirectly), making it possible for all modified to pass, but unmodified files to fail.
```
kirkrodrigues, line 172:
```suggestion
  # Note that changes to a header file may cause failures in files that include it (directly or
  # indirectly), making it possible for all modified files to pass, but unmodified files to fail.
```

kirkrodrigues (changes requested):
For the PR title, how about:

```
feat(taskfiles)!: Add filtering support to `clang-tidy` and `clang-format` tasks; Add support for running `clang-tidy` on multiple files concurrently.
```

kirkrodrigues (approved):

PR #30, opened 2025-03-07
fix(taskfile): Change `xargs --no-run-if-empty` to `xargs -I{}` to be portable between GNU and BSD's `xargs`.

davidlion:
# Description

As title says. This allows usage on macos without installing GNU utils.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

Tested with local tasks.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

- **Chores**
  - Refined the internal process for handling modified file inputs to ensure consistent execution of code quality checks, even when no files are present.

PR #31, opened 2025-03-10
fix(clang-tidy-diff): Add missing placeholder in xargs command; Set `pipefail` to detect `git diff` failures before xargs.

davidlion:
# Description

This PR contains 2 small fixes to `clang-tidy-diff`:
1. In #30 `xargs -I{} realpath` should be `xargs -I{} realpath {}`. Without the ending `{}` no argument is passed to `realpath`.
2. In a taskfile, when setting a variable with `sh` task doesn't respect a taskfile's `set` statements. Therefore, we manually set `pipefail` before calling `git diff` so that if there is an error it is returned rather than piping empty output to `xargs`.

# Checklist

  breaking change.

# Validation performed

Tested locally through CLP.

## Summary by CodeRabbit

- **Chores**
  - Updated the internal code quality checking process for C++ to enhance error handling in file modification detection, ensuring more reliable code analysis and contributing to a stable development environment.

### DIFF
--- a/taskfiles/utils/cpp-lint.yaml
+++ b/taskfiles/utils/cpp-lint.yaml
@@ -138,6 +138,8 @@ tasks:
       GIT_REF: >-
         {{default "origin/HEAD" .GIT_REF}}
       DIFF_SOURCE_FILE_PATHS_:
+        # NOTE: Although we have pipefail set globally in this file, it doesn't seem to apply here;
+        # so we need to set it again.
         sh: >-
           set -o pipefail;
           git diff --name-only --ignore-submodules "{{.GIT_REF}}" -- **/*.{cpp,h,hpp}
--- a/taskfiles/utils/cpp-lint.yaml
+++ b/taskfiles/utils/cpp-lint.yaml
@@ -141,7 +141,7 @@ tasks:
         sh: >-
           set -o pipefail;
           git diff --name-only --ignore-submodules "{{.GIT_REF}}" -- **/*.{cpp,h,hpp}
-          | xargs -I{} realpath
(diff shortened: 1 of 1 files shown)


taskfiles/utils/cpp-lint.yaml
```diff
@@ -136,11 +136,12 @@ tasks:
     label: "{{.TASK}}:{{.OUTPUT_DIR}}"
     vars:
       GIT_REF: >-
-        {{default "origin/main" .GIT_REF}}
+        {{default "origin/HEAD" .GIT_REF}}
       DIFF_SOURCE_FILE_PATHS_:
         sh: >-
```
kirkrodrigues, line 141:
```suggestion
        # NOTE: Although we have pipefail set globally in this file, it doesn't seem to apply here;
        # so we need to set it again.
        sh: >-
```

taskfiles/utils/cpp-lint.yaml
```diff
@@ -136,11 +136,12 @@ tasks:
     label: "{{.TASK}}:{{.OUTPUT_DIR}}"
     vars:
       GIT_REF: >-
-        {{default "origin/main" .GIT_REF}}
+        {{default "origin/HEAD" .GIT_REF}}
```
kirkrodrigues, line 139: Can we do this in another PR?

kirkrodrigues (changes requested):

kirkrodrigues (approved):
For the PR title, how about:

```
fix(clang-tidy-diff): Add missing placeholder in xargs command; Set pipefail to detect `git diff` failures before xargs.
```

PR #34, opened 2025-03-10
feat(lint-configs): Update the clang-format config for clang-format v20.

LinZhihao-723:
# Description

This PR updates the clang-format config to adapt release v20. Changes:
- `AllowShortNamespacesOnASingleLine`: Set to `false` to emphasize the namespace as a multiline block.
- `BreakBinaryOperations`: Set to `RespectPrecedence`. Potentially, this can make multiline condition expressions clearer (like `if (x || y || z ...)`), but I'm also okay with setting it to `Never`.
- `IndentExportBlock`: Make a similar style as `extern "C"` block.
- `KeepFormFeed`: Apparently, we don't have form feeds in our code base or plan to add ones. It should be safe to set it to `false`.
- `PenaltyBreakComment`: Set to the same value as other `PenaltyBreakXXX` options.
- `RemoveEmptyLinesInUnwrappedLines`: Set to true as it can be helpful.
- `WrapNamespaceBodyWithEmptyLines`: Set to `Never` which matches what we usually do for a namespace (no empty lines at the beginning or in the end).

The following new options are not added:
- `TemplateNames`: This should be a project-specific option.
- `VariableTemplates`: This should be a project-specific option.

Ref:
clang-format 20.1 doc: https://releases.llvm.org/20.1.0/tools/clang/docs/ClangFormatStyleOptions.html

# Checklist

  breaking change.

# Validation performed

- Using the updated config locally on clp to ensure clp core can be correctly formatted.

## Summary by CodeRabbit

- **Style**
	- Enhanced code formatting rules for C++ to enforce consistent styling, including disallowing short namespaces on a single line, preventing breaks in binary operations, and improving indentation for export blocks.
- **Chores**
	- Upgraded the required version of `clang-format` to ensure compatibility with the new formatting guidelines.

### DIFF
--- a/lint-configs/.clang-format
+++ b/lint-configs/.clang-format
@@ -27,6 +27,7 @@ AllowShortFunctionsOnASingleLine: "Inline"
 AllowShortIfStatementsOnASingleLine: "Never"
 AllowShortLambdasOnASingleLine: "All"
 AllowShortLoopsOnASingleLine: false
+AllowShortNamespacesOnASingleLine: false
 AlwaysBreakBeforeMultilineStrings: false
 BinPackArguments: false
 BinPackParameters: false
@@ -56,6 +57,7 @@ BreakBeforeBraces: "Custom"
 BreakBeforeConceptDeclarations: "Always"
 BreakBeforeInlineASMColon: "OnlyMultiline"
 BreakBeforeTernaryOperators: true
+BreakBinaryOperations: "RespectPrecedence"
 BreakConstructorInitializers: "BeforeColon"
 BreakFunctionDefinitionParameters: false
 BreakInheritanceList: "BeforeColon"
@@ -74,6 +76,7 @@ IncludeBlocks: "Regroup"
 IndentAccessModifiers: false
 IndentCaseBlocks: false
 IndentCaseLabels: true
+IndentExportBlock: true
 IndentExternBlock: "Indent"
 IndentGotoLabels: false
 IndentPPDirectives: "BeforeHash"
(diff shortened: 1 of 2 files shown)


lint-configs/.clang-format
```diff
@@ -56,6 +57,7 @@ BreakBeforeBraces: "Custom"
 BreakBeforeConceptDeclarations: "Always"
 BreakBeforeInlineASMColon: "OnlyMultiline"
 BreakBeforeTernaryOperators: true
+BreakBinaryOperations: "RespectPrecedence"
```
kirkrodrigues, line 60:
```suggestion
BreakBinaryOperations: "Never"
```

I prefer this since RespectPrecedence may look odd and regardless, we shouldn't be writing long binary expressions where precedence rules come into play without parentheses.

lint-configs/lint-requirements.txt
```diff
@@ -1,4 +1,4 @@
 # Lock to v18.x until we can upgrade our code to meet v19's formatting standards.
 clang-format~=18.1
-clang-tidy>=19.1.0
+clang-tidy>=20.1.0
```
kirkrodrigues, line 3: The modification should be for clang-format, right? And we should drop the comment as the Rabbit suggested.

kirkrodrigues (changes requested):
We also need to address the changes in the [release notes](https://releases.llvm.org/20.1.0/tools/clang/docs/ReleaseNotes.html#clang-format):

* `ReflowComments: true` -> `"Always"`

lint-configs/lint-requirements.txt
LinZhihao-723: gg, my bad

lint-configs/.clang-format
LinZhihao-723: kk, makes sense

kirkrodrigues (approved):
feat(lint-configs): Update the clang-format config for clang-format v20.

PR #40, opened 2025-04-10
feat(clang-tidy): Enforce prefixing global variables with `g_`.

davidlion:
# Description

As title says.

# Checklist

  breaking change.

# Validation performed

Ran clang-tidy on CLP.

## Summary by CodeRabbit

- **Style**
  - Updated naming standards by enforcing consistent prefixes for global pointers and variables. These configuration improvements enhance code clarity and maintainability, ensuring uniformity across the codebase. This change supports smoother development and contributes to better long-term project upkeep.

### DIFF
--- a/exports/lint-configs/.clang-tidy
+++ b/exports/lint-configs/.clang-tidy
@@ -137,7 +137,9 @@ CheckOptions:
 
   # Variable naming rules
   readability-identifier-naming.GlobalPointerCase: "lower_case"
+  readability-identifier-naming.GlobalPointerPrefix: "g_"
   readability-identifier-naming.GlobalVariableCase: "lower_case"
+  readability-identifier-naming.GlobalVariablePrefix: "g_"
   readability-identifier-naming.LocalPointerCase: "lower_case"
   readability-identifier-naming.LocalVariableCase: "lower_case"
   readability-identifier-naming.StaticVariableCase: "lower_case"


kirkrodrigues:
> Ran clang-tidy on CLP.

Did you test that global variable styles missing the prefix were detected by clang-tidy (and variables with the prefix weren't)?

kirkrodrigues (approved):
@LinZhihao-723 Any concerns with this change?

davidlion:
> > @LinZhihao-723 Any concerns with this change?
> 
> It should be good for me. Short question: shall we add `g_` for static class member variables (technically they also have global storage...)

After talking to Kirk, how about `s_` for static class member variables to set them apart (I'll update the PR)?

davidlion:
> > > @LinZhihao-723 Any concerns with this change?
> > 
> > 
> > It should be good for me. Short question: shall we add `g_` for static class member variables (technically they also have global storage...)
> 
> After talking to Kirk, how about `s_` for static class member variables to set them apart (I'll update the PR)?

After some testing, it seems clang-tidy `identifier-naming` does not differentiate static class member variables and static variables outside a class definition. The behaviour when setting `StaticVariablePrefix` seems to be:
* static class members still use `ClassMemberPrefix` (so `m_`)
* static non-class vars use `StaticVariablePrefix` (`g_`/`s_`/whatever we picked)

Proposal:
* We'll leave the `g_` setting since although ideally we never use it, if we ever do it is better to have the setting than not.
* We won't set `StaticVariablePrefix` since it would be confusing having different prefixes for class/non-class static variables.

LinZhihao-723:
> > > > @LinZhihao-723 Any concerns with this change?
> > > 
> > > 
> > > It should be good for me. Short question: shall we add `g_` for static class member variables (technically they also have global storage...)
> > 
> > 
> > After talking to Kirk, how about `s_` for static class member variables to set them apart (I'll update the PR)?
> 
> After some testing, it seems clang-tidy `identifier-naming` does not differentiate static class member variables and static variables outside a class definition. The behaviour when setting `StaticVariablePrefix` seems to be:
> 
> * static class members still use `ClassMemberPrefix` (so `m_`)
> * static non-class vars use `StaticVariablePrefix` (`g_`/`s_`/whatever we picked)
> 
> Proposal:
> 
> * We'll leave the `g_` setting since although ideally we never use it, if we ever do it is better to have the setting than not.
> * We won't set `StaticVariablePrefix` since it would be confusing having different prefixes for class/non-class static variables.

Ok I see. The proposal lgtm. We can keep the `g_` setting introduced in this PR. For static class members, if there's no way to elegantly enforce `s_`, I'm ok to leave it `m_`.

LinZhihao-723 (approved):

PR #42, opened 2025-04-25
feat(taskfiles): Add tasks to install a list of boost libraries as dependencies.

sitaowang1998:
# Description

`Boost` does not support build with `cmake`, so we cannot use the tasks from `cmake.yaml` to install `Boost`.

This PR adds:
* A taskfile to download and install `Boost`.
* A taskfile that uses the new task to install `Boost`.
* A simple `C++` project that uses the installed `Boost` library as dependency.
* A task to run the whole test, and call it in main taskfile in tests.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

* [x] `task tests:boost` ran successfully.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

## Summary by CodeRabbit

- **New Features**
  - Added automated tasks for downloading, building, and installing Boost C++ libraries.
  - Introduced Boost-specific build and test configurations, including a CMake setup and formatting rules.
  - Implemented comprehensive Boost test suite covering multiple Boost components.
  - Added automated tests for remote zip file download and extraction with inclusion and exclusion filters.

- **Refactor**
  - Streamlined test orchestration by consolidating and delegating test tasks to dedicated taskfiles.

- **Chores**
  - Updated main test task to run all tests using a unified command.

PR #47, opened 2025-05-07
fix(lint-configs): Relax clang-tidy `misc-include-cleaner` check to a warning.

davidlion:
# Description

Many C++ libraries expose a single "helper/meta" header that contains many other "internal" headers. This creates issues with `iwyu` style checkers like clang-tidy's `misc-include-cleaner` as these "meta" headers do not directly expose the symbols used. Previously we have gotten around this by creating [facade headers in ystdlb](https://github.com/y-scope/ystdlib-cpp/tree/main/src/ystdlib/wrapped_facade_headers), but this problem is becoming hard to keep up with.

This PR addresses this problem by relaxing `misc-include-cleaner` back to a warning so the issues will not block CI. 

# Checklist

  breaking change.

# Validation performed

Tested on boost test code in #42.

## Summary by CodeRabbit

- **Chores**
  - Updated linting configuration to exclude a specific warning from being treated as an error.

### DIFF
--- a/exports/lint-configs/.clang-tidy
+++ b/exports/lint-configs/.clang-tidy
@@ -1,5 +1,11 @@
 FormatStyle: "file"
-WarningsAsErrors: "*,-misc-include-cleaner"
+
+# Left as warnings:
+# - `misc-include-cleaner` because it's difficult to resolve iwyu issues for external libraries
+# without facade headers.
+WarningsAsErrors: >-
+  *
+  -misc-include-cleaner
 
 # Disabled checks:
 # - `bugprone-easily-swappable-parameters` because it's difficult to mitigate.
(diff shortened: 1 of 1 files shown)


exports/lint-configs/.clang-tidy
```diff
@@ -1,5 +1,5 @@
 FormatStyle: "file"
-WarningsAsErrors: "*"
+WarningsAsErrors: "*,-misc-include-cleaner"
```
kirkrodrigues, line 2: Do you want to use the same format as `Checks` for readability?

kirkrodrigues (changes requested):

kirkrodrigues (approved):
For the PR title, how about:

```
fix(lint-configs): Relax clang-tidy `misc-include-cleaner` check to a warning.
```

PR #54, opened 2025-05-14
ci: Add GH workflows to validate PR titles to follow Conventional Commits and run tests.

PacificViking:
# Description

Add GitHub automated testing using `task` and PR title checking from https://github.com/y-scope/ystdlib-cpp/pull/6.

Note: Tests were not run on macOS GH runners because of issue https://github.com/y-scope/yscope-dev-utils/issues/52 causing them to fail.

# Checklist

  breaking change.

# Validation performed

Workflow automated testing occurs and passes on my forked repository. Output matches tests being run from my local machine.

## Summary by CodeRabbit

- **Chores**
  - Introduced automated checks to enforce semantic pull request titles.
  - Added automated unit test workflow running on multiple Ubuntu versions for improved reliability.

### DIFF
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -29,10 +29,6 @@ jobs:
         with:
           submodules: "recursive"
 
-      - uses: actions/setup-node@v4
-        with:
-          node-version: "18"
-
       - name: "Install task"
         shell: "bash"
         run: "npm install -g @go-task/cli"
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -29,9 +29,9 @@ jobs:
         with:
           submodules: "recursive"
 
-      - uses: "actions/setup-python@v5"
+      - uses: actions/setup-node@v4
         with:
-          python-version: "3.11"
+          node-version: "18"
 
       - name: "Install task"
         shell: "bash"
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -1,0 +1,41 @@
+name: "unit-tests"
+
+on:
+  pull_request:
+  push:
+  schedule:
+    # Run daily at 00:20 UTC
+    - cron: "20 0 * * *"
+  workflow_dispatch:
+
+permissions: {}
+
+concurrency:
+  group: "${{github.workflow}}-${{github.ref}}"
+
+  # Cancel in-progress jobs for efficiency
+  cancel-in-progress: true
+
+jobs:
+  build-ystdlib-cpp:
+    strategy:
+      matrix:
+        os:
+          - "ubuntu-22.04"
+          - "ubuntu-24.04"
+    runs-on: "${{matrix.os}}"
+    steps:
+      - uses: "actions/checkout@v4"
+        with:
+          submodules: "recursive"
+
+      - uses: "actions/setup-python@v5"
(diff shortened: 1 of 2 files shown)


.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,41 @@
...
+      matrix:
+        os:
+          - "ubuntu-22.04"
+          - "ubuntu-24.04"
+    runs-on: "${{matrix.os}}"
+    steps:
+      - uses: "actions/checkout@v4"
+        with:
+          submodules: "recursive"
+
```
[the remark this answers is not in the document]
PacificViking: This is fixed in the latest push.

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,41 @@
...
+on:
+  pull_request:
+  push:
+  schedule:
+    # Run daily at 00:20 UTC
+    - cron: "20 0 * * *"
+  workflow_dispatch:
+
+permissions: {}
```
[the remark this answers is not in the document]
PacificViking: actions/checkout did not fail on my forked repository.

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,41 @@
...
+  group: "${{github.workflow}}-${{github.ref}}"
+
+  # Cancel in-progress jobs for efficiency
+  cancel-in-progress: true
+
+jobs:
+  unit-tests:
+    strategy:
+      matrix:
+        os:
```
kirkrodrigues, line 23:
```suggestion
        os:
          - "macos-14"
          - "macos-14"
```

Unless there are issues with running on macOS? The reason to test on macOS is that some developers use macOS.

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,41 @@
...
+          - "ubuntu-24.04"
+    runs-on: "${{matrix.os}}"
+    steps:
+      - uses: "actions/checkout@v4"
+        with:
+          submodules: "recursive"
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "18"
```
kirkrodrigues, line 34: All GitHub runners we use should have `npm` installed, so I don't think we need this. I agree reproducibility is important, but since we're just using `npm` as a package manager, I don't think it will have a meaningful impact to use a specific node version (unless we're talking about older versions). What do you think?

kirkrodrigues (changes requested):

.github/workflows/unit-tests.yaml
PacificViking: There are issues with running on MacOS. The boost test cases have a dependency on /dev/true which is only available on some linux distributions.

.github/workflows/unit-tests.yaml
PacificViking: Makes sense. My new push reflects this.

.github/workflows/unit-tests.yaml
kirkrodrigues: Ah gotcha. Can you add that to the PR description and link tot he issue you opened? That way if someone else comes back to this PR/change later, they know why we didn't add macOS.

kirkrodrigues (approved):
For the PR title, how about:

```
ci: Add GH workflows to validate PR titles follow Conventional Commits and to run tests.
```

PR #59, opened 2025-06-03
feat(taskfiles)!: Add python script to replace confusing shell find usage in tasks.

davidlion:
# Description

Previously, many tasks used the shell command `find` to search and filter files to work on. This resulted in a difficult to read combination of shell, yaml, and go templating. This PR replaces the usage of shell find with a python script making the tasks more readable.

Unit tests for the find script are also added.

# Checklist

* [x] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [x] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [x] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

Tested with new unit tests and locally with ffi-go.

[yscope-contrib-guidelines]: https://docs.yscope.com/dev-guide/contrib-guides-overview.html

PR #60, opened 2025-06-03
ci: Add taskfiles and update GH workflows for linting Python files.

davidlion:
# Description

As title says. ~This PR depends on #59.~

~Currently, there are some known warnings due to the ruff configuration having incompatible settings from enabling all linting checks. I'll investigate these issues, but they don't block the rest of the review.~
These warnings have been resolved and changes made to `ruff.toml`.

`exports/lint-configs/python/mypy.ini` and `exports/lint-configs/python/ruff.toml` settings are duplicated from `exports/ystdlib-py/pyproject.toml` as there isn't a way to cleanly inherit the options. Changes to these files will need to be copied into pyproject files, but it is still useful to provide a clear standalone version of the recommended configuration.

[`uv tool`](https://docs.astral.sh/uv/guides/tools/) was not used as a solution as (at the time of the PR) installation is scoped to the user-level making it awkward to have different versions of the same tool for different projects. One work around would be to set an [environment variable](https://docs.astral.sh/uv/reference/environment/#uv_tool_dir) on every use to point to project specific tool directories/installations, but this is just a more confusing (and less powerful) solution compared to using [`uv run --project`](https://docs.astral.sh/uv/reference/cli/#uv-run--project). Another option would be to somehow add the correct project and tool specific [`--from` flag](https://docs.astral.sh/uv/guides/tools/#requesting-specific-versions) to every call, but that is also more confusing/complicated to implement than the `--project` flag. Additionally, neither of these workarounds are automatically picked up when running `uv` locally, whereas `uv` will find and use local pyproject files.

# Checklist

  breaking change.

# Validation performed

Tasks passing.

## Summary by CodeRabbit

- **New Features**
  - Introduced Python linting and type checking with Ruff and Mypy, including strict mode and enhanced error reporting.
  - Added configuration files and documentation for Python linting and type checking.
- **Chores**
  - Updated automation workflows to include Python files in lint checks and to install required tools automatically.
  - Improved task automation by introducing new variables for directory paths and commands.
  - Streamlined taskfile configurations for better maintainability and path consistency.


exports/lint-configs/python/ruff.toml
```diff
@@ -0,0 +1,17 @@
...
+select = ["ALL"]
+ignore = [
+    # Redundant and conflicts with ruff format
+    "COM812",
+    # No blank line before docstrings (D211)
+    "D203",
+    # Docstring summary line on the next line after quotes (D213)
+    "D212",
+]
+isort.order-by-type = false
```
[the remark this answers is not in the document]
davidlion: https://docs.astral.sh/ruff/settings/#lint_isort_order-by-type shows it uses hyphens.

exports/lint-configs/python/mypy.ini
```diff
@@ -0,0 +1,4 @@
+[mypy]
+pretty = true
+strict = true
+warn_unused_ignores = true
```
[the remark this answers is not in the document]
davidlion: These are all set using `strict = true`.

taskfile.yaml
```diff
@@ -8,7 +8,11 @@ set: ["u", "pipefail"]
 shopt: ["globstar"]
 
 vars:
+  G_EXPORTS_DIR: "{{.ROOT_DIR}}/exports"
+  G_LINT_CONFIGS_DIR: "{{.G_EXPORTS_DIR}}/lint-configs"
   G_OUTPUT_DIR: "{{.ROOT_DIR}}/build"
+  G_YSTDLIB_PY_DIR: "{{.G_EXPORTS_DIR}}/ystdlib-py"
+  G_PYFIND_CMD: "uv run --project '{{.G_YSTDLIB_PY_DIR}}' pyfind"
```
LinZhihao-723, line 15:
```suggestion
  G_PYFIND_CMD: "uv run --project '{{.G_YSTDLIB_PY_DIR}}' pyfind"
  G_YSTDLIB_PY_DIR: "{{.G_EXPORTS_DIR}}/ystdlib-py"
```
Assuming we order these variables in alphabetical order.

taskfiles/lint/python.yaml
```diff
@@ -0,0 +1,22 @@
...
+    cmds:
+      - task: "py"
+        vars:
+          RUFF_CHECK_FLAGS: "--fix"
+
+  py:
+    internal: true
+    cmd: |-
+      cd "{{.G_YSTDLIB_PY_DIR}}"
+      uv run mypy {{.MYPY_FLAGS}} "./src"
```
LinZhihao-723, line 20: Is this defined?

LinZhihao-723 (changes requested):
Overall lgtm. One suggestion: as explained in the PR description

> exports/lint-configs/python/mypy.ini and exports/lint-configs/python/ruff.toml settings are duplicated from exports/ystdlib-py/pyproject.toml as there isn't a way to cleanly inherit the options. Changes to these files will need to be copied into pyproject files, but it is still useful to provide a clear standalone version of the recommended configuration.

It might be worth to have a README in `exports/lint-configs/python` to explain what these files are and how to use them (by actually copying the content with necessary overrides).

taskfile.yaml
davidlion: Iirc this breaks since `G_PYFIND_CMD` relies on `G_YSTDLIB_PY_DIR`. I'll try spliting `G_PYFIND_CMD` into another grouping/category to make it nicer.

davidlion:
> Overall lgtm. One suggestion: as explained in the PR description
> 
> > exports/lint-configs/python/mypy.ini and exports/lint-configs/python/ruff.toml settings are duplicated from exports/ystdlib-py/pyproject.toml as there isn't a way to cleanly inherit the options. Changes to these files will need to be copied into pyproject files, but it is still useful to provide a clear standalone version of the recommended configuration.
> 
> It might be worth to have a README in `exports/lint-configs/python` to explain what these files are and how to use them (by actually copying the content with necessary overrides).

Good point.

taskfiles/lint/python.yaml
LinZhihao-723: yeah let's try to remove it for now then. We may want this to be less confusing as other Python projects will probably try to copy the setup from this repo before we have reusable task utils.

exports/lint-configs/python/README.md
```diff
@@ -0,0 +1,9 @@
+# Python linter configs
+
+This directory contains standalone Python linter configuration files that double as a source of
+reference settings for using the linters in your project.
+
+For a project with its own `pyproject.toml` file, at the time of writing, there is not a clean,
+general solution to include the standalone configuration files for different linters. In this
+scenario, the settings inside the configuration files should be copied into your `pyproject.toml`
+file, as seen in [ystdlib-py/pyproject.toml](../../ystdlib-py/pyproject.toml).
```
LinZhihao-723, line 9:
```suggestion
For a project with its own `pyproject.toml` file, at the time of writing, there is no clean, general
solution for including the standalone configuration files for different linters. In this scenario,
the settings inside the configuration files should be copied into your `pyproject.toml` file, as
seen in [ystdlib-py/pyproject.toml](../../ystdlib-py/pyproject.toml).
```
Minor:
`there is not a clean` -> `there is no clean`
`to include` -> `for including`

LinZhihao-723 (changes requested):
one minor suggestion otherwise lgtm. I will approve after this is fixed.

PR #62, opened 2025-06-07
fix(ystdlib-py)!: Rename `find.py` to `pyfind.py` to avoid shadowing `find`.

davidlion:
# Description

Previously, when the virtual env was active ystdlib-py's find script would shadow the normal `find` utility. This is because the script in `pyproject.toml` is also named `find` and installed into the virtual env as such by `uv`.
To avoid this conflict we rename everything to `pyfind` and `pyfind.py`.

# Checklist

  breaking change.

# Validation performed

All tests still passing.


kirkrodrigues (changes requested):
* The docstring at the top of `pyfind.py` needs to be updated.
* We should probably rename the test taskfile (`test-find.yaml`) and output directory (`tests-find`), right?

kirkrodrigues (approved):

PR #80, opened 2025-08-23
feat(taskfiles)!: Add `checksum` header comment; Update and refactor Boost tasks by:

davidlion:
* Use checksum tasks to prevent unnecessary repetition of Boost tasks.
* Remove quotes and spaces from Boost parameter strings.
* Remove unused GENERATE_* arguments.
* Update Boost unit tests to use 1.89.0.

# Description

(Depends on #81.)

Previously, we avoided using taskfile `sources` and `generates` with build systems (CMake and Boost's b2) and deferred to them to decide what work to repeat. This has been somewhat acceptable for CMake projects, but causes long build times for projects using Boost.

This PR modifies the Boost tasks to use the existing checksum tasks to mark successful states of the build process, allowing future builds to skip running unnecessary tasks. A detailed header comment is added to `exports/taskfiles/utils/checksum.yaml` detailing the recommended usage.

In addition it makes a few small changes:
* Remove quotes and spaces from Boost parameter strings, which would cause them to not work correctly.
* Removes unused GENERATE_* arguments in preference of fixed versions.
* Updates the Boost unit tests to use Boost 1.89.

## Additional comments on updating checksums

Another approach considered, was to "snapshot" the state at the end of each task. This would be done by having each task copy the contents of its sources to a new location before mutating them. The main issue with this approach is that if any of these contents stored a path, the path would be made incorrect by the copy. This is not acceptable in many cases and prevents it from being a "general" solution. However, this approach may still be preferable in cases where it is important to recover from absolutely any stage.

# Checklist

  breaking change.

# Validation performed

Tested locally using the Boost unit tests and ystdlib-cpp by deleting certain fails and observing what tasks ran.
Added some unit tests for the checksum tasks to verify behaviour.

## Summary by CodeRabbit

* **Chores**
  * Upgraded Boost from 1.86.0 to 1.89.0.
  * Removed Boost.System from test dependencies.

* **Refactor**
  * Restructured download/build/install flows to a tar-based, single-source model with unified source/build/install directories.
  * Added per-stage checksum and sentinel artifacts for stronger artifact validation.

* **Chores**
  * Enabled stricter shell error handling and recursive globbing in checksum tooling.


exports/taskfiles/utils/boost.yaml
```diff
@@ -8,38 +8,39 @@ set: ["u", "pipefail"]
...
   generate:
     internal: true
+    label: "{{.TASK}}:{{.SOURCE_DIR}}"
     vars:
+      CHECKSUM_FILE: >-
+        {{default (printf "%s.md5" .SOURCE_DIR) .CHECKSUM_FILE}}
       EXTRA_ARGS:
         ref: "default (list) .EXTRA_ARGS"
+      SOURCE_SENTINEL_FILE: "{{.SOURCE_DIR}}/{{.TASK}}.sentinel"
     requires:
```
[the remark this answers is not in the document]
davidlion: This type of sentinel file is only added to a directory when the task needs to alter the contents of the directory. Calling this task would be a logical bug if the upstream workflow required the directory to remain unchanged.

exports/taskfiles/utils/checksum.yaml
```diff
@@ -1,13 +1,45 @@
...
+# the contents still match the checksum in later runs.
+#
+# When creating a checksum a task should:
+#
+# a. Call the `compute` task at the end to create the checksum file.
+# b. Call `validate` in `deps` to verify the checksum file matches the current contents.
+#   - `validate` will delete the checksum file if there is no match.
+# c. Add the checksum file to `generates`. This ensures the task runs if:
+#   - The task has not run before.
+#   - `validate` deleted the checksum file as the contents did not match.
```
Bill-hbrhbr, line 19:
```suggestion
# c. Add the checksum file to `generates`. This ensures the task runs if:
#   - The task has not run before.
#   - `validate` deleted the checksum file as the contents did not match.
# d. Add the checksum file to the `sources` of any task that depends on it, so the downstream task
#    reruns when the checksum is invalidated and recomputed.
```

exports/taskfiles/utils/boost.yaml
```diff
@@ -8,36 +8,39 @@ set: ["u", "pipefail"]
...
+        {{default (printf "%s.md5" .SOURCE_DIR) .CHECKSUM_FILE}}
+      SOURCE_SENTINEL_FILE: "{{.SOURCE_DIR}}/{{.TASK}}.sentinel"
     requires:
-      vars: ["SOURCE_DIR", "GENERATE_DIR", "GENERATE_CHECKSUM_FILE", "INSTALL_PREFIX", "TARGETS"]
-    sources:
-      - "{{.SOURCE_DIR}}/**/*"
+      vars:
+        - "INSTALL_PREFIX"
+        - "SOURCE_DIR"
+        - "TARGETS"
```
Bill-hbrhbr, line 32: we should dump these input vars into a config file and cite it as a `source` of the **generate** stage.
But if we do the dumping inside the current **generate** task:
- first run: no config file
- second run: sees config file
- third run: sees config file does not change and thus up-to-date
The **generate** stage will run twice before detecting up-to-date status. So we should put the config dumping into a standalone task that runs in the `deps` section of **boost:generate**.

Bill-hbrhbr: The `label: "{{.TASK}}:{{.SOURCE_DIR}}"` looks sound.

Bill-hbrhbr (changes requested):

exports/taskfiles/utils/boost.yaml
```diff
@@ -8,36 +8,39 @@ set: ["u", "pipefail"]
...
-  # Generates `GENERATE_DIR` by copying `SOURCE_DIR` and then running boost's bootstrap step.
+  # Runs Boost's generate (bootstrap) step.
+  # A sentinel file is written to `SOURCE_DIR` and added to `generates` to explicitly track whether
+  # this task updated `SOURCE_DIR`.
   #
-  # @param {string} SOURCE_DIR Project source directory.
-  # @param {string} GENERATE_DIR Directory in which to generate `b2` and the build configuration.
-  # @param {string} GENERATE_CHECKSUM_FILE Checksum file for `GENERATE_DIR`.
   # @param {string} INSTALL_PREFIX Path prefix of where the project should be installed.
+  # @param {string} SOURCE_DIR Project source directory.
```
Bill-hbrhbr, line 17: I don't know if we should change all `SOURCE_DIR` to `EXTRACTION_DIR`. The latter is consistent with the terminology used for regular cmake libs.

exports/taskfiles/utils/boost.yaml
davidlion: Rather than a config file, how about we just add all the input variables to the label so that task handles different input naturally? My original change should really have done that to start with and it is pretty buggy to ignore input var changes.

The only benefit of a config file I can think of is that we might be able to add logic where certain input changes do not redo all work, but that seems like a relatively small benefit.

davidlion: I think there still might be some weird cases if two invocations share SOURCE_DIR. I think a config file would fix that case if the config file is included in the checksum.

I think adding all the input variables to the checksum name (basically make the checksum name the label) might fix things. This way even if two invocations have overlapping input they'd generate their own separate checksum files. If one is run again it would only be skipped if the contents actually match its own checksum file (it wouldn't mistakenly check against the checksum from an invocation with different input). (Pretty sure the sentinel file would need a similar name change.)

PR #81, opened 2025-08-26
fix(taskfiles): Output unexpected errors in checksum tasks; Call `gtar` on macOS (fixes #67); Add unit tests for checksum tasks.

### LINKED ISSUE #67: Docs: missing documentation of macOS `gtar` requirement
### Bug

as the title suggests, we should have some docs to say `gtar` should be installed, as that's been required since #21 

### yscope-dev-utils version

92cc1f3661e56e07fe165c60b74ea3910efaf097

### Environment

```
junhao@Junhaos-MBP ~ % sw_vers 
ProductName:		macOS
ProductVersion:		15.5
BuildVersion:		24F74
```

### Reproduction steps

This was found as running `task deps` in y-scope/clp-ffi-js@13c31edef9569b678d4905d6512e87fdfc4589dd on mac fails.

```
# even with "--verbose", the cause of the command failure isn't very clear
junhao@1498777c24d1 clp-ffi-js % task deps --verbose
...
task: [utils:remote:download-and-extract-tar-/Users/junhao/workspace/clp-ffi-js/build/deps/ystdlib-src] gtar --extract --strip-components="1" --directory "/Users/junhao/workspace/clp-ffi-js/build/deps/ystdlib-src" --file "/Users/junhao/workspace/clp-ffi-js/build/deps/ystdlib-src.tar.gz" --no-anchored --wildcards 2> /dev/null
task: Failed to run task "deps:default": exit status 127
```

davidlion:
# Description

This PR addresses a few issues with the `checksum` tasks and also adds unit tests for these tasks.

1. `checksum:compute` requires gnu tar for its flags and would fail silently on MacOS (unless `tar` was set to gnu tar)
     * Now calls `gtar` if running on macos.
2. Both compute and validate have use cases where they are expected to fail leading to the tasks to always succeed unconditionally. This is confusing/problematic if there actually is an error.
    * Both tasks now have a `FAIL` flag allowing users to change whether they wish to allow the task to fail or not.
    * Both tasks now record their errors and print them if a failure was unexpected.

For `compute`, `FAIL` defaults to `true` as there is currently only one case where we do not want to fail. When we call `compute` from `validate` we want task to continue if `compute` fails as `validate` will delete the checksum file so that the calling task will not be skipped (assuming it correctly includes the checksum in its `generates` field). Previously, `compute` used `ignore_error:true` to blindly succeed in all cases.

For `validate`, `FAIL` defaults to `false` as generally the "failing" workflow is for `validate` to delete the checksum so that the caller runs and (re-)creates all the content for the checksum. In the new workflow, where we "snapshot" different stages in a build by computing checksums at each stage, a caller may need to validate a checksum which it will not (re-)create the content of (e.g. a `build` task will want to validate the checksum from `generate`, but `build` is not capable of (re-)running `generate`). In this case, `validate` should cause task to fail as the caller has no way to recover.

# Checklist

  breaking change.

# Validation performed

* New unit tests are passing.
* Tested #80 with the changes.
* Some manual verification of failing cases.

## Summary by CodeRabbit

* **Documentation**
  * Added macOS prerequisites noting missing GNU utilities and recommended brew packages, with reference links.

* **New Features**
  * Configurable fail‑soft / fail‑fast controls for checksum operations, improved error logging, path-existence checks, streaming checksum handling and robust cleanup behaviour.

* **Tests**
  * Added automated checksum test suite (rerun, skip, update scenarios) and integrated it into the test task group.

### DIFF
--- a/exports/taskfiles/utils/checksum.yaml
+++ b/exports/taskfiles/utils/checksum.yaml
@@ -10,8 +10,8 @@ tasks:
   #
   # @param {string} CHECKSUM_FILE
   # @param {string[]} INCLUDE_PATTERNS Path wildcard patterns to compute the checksum for.
-  # @param {string[]} [EXCLUDE_PATTERNS=[]] Path wildcard patterns, relative to any `INCLUDE_PATTERNS`,
-  # to exclude from the checksum.
+  # @param {string[]} [EXCLUDE_PATTERNS=[]] Path wildcard patterns, relative to any
+  # `INCLUDE_PATTERNS`, to exclude from the checksum.
   # @param {string} [IGNORE_ERROR="false"] If set to "true", the task will not fail on error.
   compute:
     internal: true
@@ -57,7 +57,7 @@ tasks:
           {{- range .INCLUDE_PATTERNS}}
           {{.}} \
           {{- end}}
-          2> "{{.TMP_ERR_LOG}}" \
+          2>> "{{.TMP_ERR_LOG}}" \
             | md5sum > "{{.CHECKSUM_FILE}}" \
         ; then
           rm -f "{{.CHECKSUM_FILE}}"
@@ -75,8 +75,8 @@ tasks:
   #
   # @param {string} CHECKSUM_FILE
   # @param {string[]} INCLUDE_PATTERNS Path wildcard patterns to validate the checksum for.
-  # @param {string[]} [EXCLUDE_PATTERNS=[]] Path wildcard patterns, relative to any `INCLUDE_PATTERNS`,
-  # to exclude from the checksum.
+  # @param {string[]} [EXCLUDE_PATTERNS=[]] Path wildcard patterns, relative to any
+  # `INCLUDE_PATTERNS`, to exclude from the checksum.
   # @param {string} [FAIL_ON_ERROR="false"] If set to "true", the task fails when checksums
   # mismatch.
   validate:
@@ -118,7 +118,7 @@ tasks:
               cmp -s "{{.TMP_CHECKSUM_FILE}}" "{{.CHECKSUM_FILE}}" \
               || (
                 echo "cmp failed for '{{.TMP_CHECKSUM_FILE}}' '{{.CHECKSUM_FILE}}'" \
-                  > "{{.TMP_ERR_LOG}}"
+                  >> "{{.TMP_ERR_LOG}}"
                 exit 1
                 )
             )


exports/taskfiles/utils/checksum.yaml
```diff
@@ -24,58 +35,96 @@ tasks:
...
+          CHECKSUM_FILE: "{{.TMP_CHECKSUM_FILE}}"
           INCLUDE_PATTERNS:
             ref: ".INCLUDE_PATTERNS"
           EXCLUDE_PATTERNS:
-            ref: "default (list) .EXCLUDE_PATTERNS"
-          CHECKSUM_FILE: "{{.TMP_CHECKSUM_FILE}}"
-      - defer: "rm -f '{{.TMP_CHECKSUM_FILE}}'"
-      # Check that all paths exist and the checksum matches; otherwise delete the checksum file.
+            ref: ".EXCLUDE_PATTERNS"
+          FAIL: "false"
```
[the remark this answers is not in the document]
davidlion: Sorry, the motivation is confusing since I split the PRs. This is necessary for better checksum workflows like in boost. I'll improve the description.
https://github.com/y-scope/yscope-dev-utils/pull/80/files#diff-a1f9d497b81cb7bdb53e553d8fce46fbee451bbe4629106955d2536ada79015dR116

In this case when we validate the source directory and it fails we want to completely fail.
One option is to just assume this cannot fail and not check it, but that seems a bit much to avoid a flag in rare cases.
The other option is to try and recover, but that is very complicated. We'd need `boost:build-and-install` to somehow re-run `boost:generate`, which is a bit ugly/complicated. More generally any tasks that implicitly depends on another task (but cannot be `deps` without a lot of work due to their library/helper style) will need to support a recovery pass in certain cases.

taskfiles/checksum/tests.yaml
```diff
@@ -0,0 +1,148 @@
+version: "3"
+
+includes:
+  checksum:
+    internal: true
+    taskfile: "../../exports/taskfiles/utils/checksum.yaml"
+
+tasks:
+  default:
```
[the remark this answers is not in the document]
davidlion: I agree with you and want to, but afaik we cannot actually test if a task fails using task itself. E.g. if `FAIL` is `true` and the task fails, the entire task "run" will stop and exit.
Essentially, (afaik) there is no way to write a unit test (in task) that verifies a task fails.

We probably need to switch to using python to test tasks eventually.

exports/taskfiles/utils/checksum.yaml
```diff
@@ -1,20 +1,31 @@
 version: "3"
 
-set: ["u", "pipefail"]
-shopt: ["globstar"]
-
 tasks:
+
```
Bill-hbrhbr:
```suggestion
```

Bill-hbrhbr (dismissed):
lgtm

exports/taskfiles/utils/checksum.yaml
```diff
@@ -25,58 +36,97 @@ tasks:
...
+            {{- end}}
+            {{- range .INCLUDE_PATTERNS}}
+            {{.}} \
+            {{- end}}
+            2> "{{.TMP_ERR_LOG}}" \
+                | md5sum > "{{.CHECKSUM_FILE}}" \
+        ; then
+          rm -f "{{.CHECKSUM_FILE}}"
+          {{- if eq "true" .FAIL}}
+            printf "[{{.TASK}} error] failed with:\n%s\n" "$(cat {{.TMP_ERR_LOG}})"
```
junhaoliao:
```suggestion
            printf "[{{.TASK}} error] failed with:\n%s\n" "$(cat '{{.TMP_ERR_LOG}}')"
```

exports/taskfiles/utils/checksum.yaml
```diff
@@ -25,58 +36,97 @@ tasks:
...
+  # reference checksum stored in the given file. If validation fails, the checksum file is deleted,
+  # but the task succeeds so dependent tasks that list the checksum file under `generates` will
+  # rerun automatically. The calling task can set `FAIL` to "true" to make this task fail when the
+  # validation is expected to succeed.
+  #
   # @param {string} CHECKSUM_FILE
   # @param {string[]} INCLUDE_PATTERNS Path wildcard patterns to validate the checksum for.
   # @param {string[]} [EXCLUDE_PATTERNS] Path wildcard patterns, relative to any `INCLUDE_PATTERNS`,
   # to exclude from the checksum.
+  # @param {string} [FAIL="false"] If set to "true", the task fails when checksums mismatch.
```
junhaoliao: would `FAIL_ON_ERROR` be more self-explanatory?

exports/taskfiles/utils/checksum.yaml
```diff
@@ -25,58 +36,97 @@ tasks:
...
+            --no-anchored \
+            --wildcards \
+            {{- range .EXCLUDE_PATTERNS}}
+            --exclude="{{.}}" \
+            {{- end}}
+            {{- range .INCLUDE_PATTERNS}}
+            {{.}} \
+            {{- end}}
+            2> "{{.TMP_ERR_LOG}}" \
+                | md5sum > "{{.CHECKSUM_FILE}}" \
```
junhaoliao: indent
```suggestion
          {{.ARCHIVER}} \
          --create \
          --file - \
          --group 0 \
          --mtime "UTC 1970-01-01" \
          --numeric-owner \
          --owner 0 \
          --sort name \
          --no-anchored \
          --wildcards \
          {{- range .EXCLUDE_PATTERNS}}
          --exclude="{{.}}" \
          {{- end}}
          {{- range .INCLUDE_PATTERNS}}
          {{.}} \
          {{- end}}
          2> "{{.TMP_ERR_LOG}}" \
            | md5sum > "{{.CHECKSUM_FILE}}" \
```

exports/taskfiles/utils/checksum.yaml
```diff
@@ -25,58 +36,97 @@ tasks:
...
-          diff -q "{{.TMP_CHECKSUM_FILE}}" "{{.CHECKSUM_FILE}}" 2> /dev/null
-        ) || rm -f "{{.CHECKSUM_FILE}}"
+              (
+                cmp -s "{{.TMP_CHECKSUM_FILE}}" "{{.CHECKSUM_FILE}}" \
+                || (
+                  echo "cmp failed for '{{.TMP_CHECKSUM_FILE}}' '{{.CHECKSUM_FILE}}'" \
+                    > "{{.TMP_ERR_LOG}}"
+                  exit 1
+                  )
+              )
```
junhaoliao:
```suggestion
            (
              cmp -s "{{.TMP_CHECKSUM_FILE}}" "{{.CHECKSUM_FILE}}" \
              || (
                echo "cmp failed for '{{.TMP_CHECKSUM_FILE}}' '{{.CHECKSUM_FILE}}'" \
                  > "{{.TMP_ERR_LOG}}"
                exit 1
                )
            )
```

exports/taskfiles/utils/checksum.yaml
```diff
@@ -25,58 +36,97 @@ tasks:
...
+        ) \
+            || (
+              {{- if eq "true" .FAIL}}
+                printf "[{{.TASK}} error] failed with:\n%s\n" "$(cat {{.TMP_ERR_LOG}})"
+                exit 1
+              {{- else}}
+                rm -f "{{.CHECKSUM_FILE}}"
+              {{- end}}
+            )
```
junhaoliao:
```suggestion
          || (
            {{- if eq "true" .FAIL}}
              printf "[{{.TASK}} error] failed with:\n%s\n" "$(cat {{.TMP_ERR_LOG}})"
              exit 1
            {{- else}}
              rm -f "{{.CHECKSUM_FILE}}"
            {{- end}}
          )
```

exports/taskfiles/utils/checksum.yaml
junhaoliao: or maybe the 4-space indents are intentional? i can't be sure what are the conventions when we mix shell commands with go templates...

the code is fairly readable either way anyways

exports/taskfiles/utils/checksum.yaml
```diff
@@ -1,20 +1,30 @@
...
 tasks:
+  # Compute the checksum of the given path include/exclude patterns, saving the result to
+  # `CHECKSUM_FILE`. The calling task can set `IGNORE_ERROR` to "false" if they wish to continue if
+  # checksum computation fails.
+  #
   # @param {string} CHECKSUM_FILE
   # @param {string[]} INCLUDE_PATTERNS Path wildcard patterns to compute the checksum for.
   # @param {string[]} [EXCLUDE_PATTERNS] Path wildcard patterns, relative to any `INCLUDE_PATTERNS`,
   # to exclude from the checksum.
+  # @param {string} [IGNORE_ERROR="false"] If set to "true", the task will not fail on error.
```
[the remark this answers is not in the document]
davidlion: Can you re-write this in english please? I failed french.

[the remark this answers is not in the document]
davidlion: 1. The `[IGNORE_ERROR="false"]` means that the variable defaults to false, but the comment after states `If set to "true", the task will not fail on error.`.
2. Task only respects/sets shell flags like pipefail from the root taskfile, which will never be `exports/taskfiles/utils/checksum.yaml`. However, I will add a note to the doc string that the task will not behave correctly if pipefail isn't set in the root task file.

exports/taskfiles/utils/checksum.yaml
```diff
@@ -25,58 +38,95 @@ tasks:
...
 
+  # Validates that the checksum computed from the given include/exclude path patterns matches the
+  # reference checksum stored in the given file. If validation fails, the checksum file is deleted,
+  # but the task succeeds so dependent tasks that list the checksum file under `generates` will
+  # rerun automatically. The calling task can set `FAIL_ON_ERROR` to "true" to make this task fail
+  # when the validation is expected to succeed.
+  #
   # @param {string} CHECKSUM_FILE
   # @param {string[]} INCLUDE_PATTERNS Path wildcard patterns to validate the checksum for.
   # @param {string[]} [EXCLUDE_PATTERNS] Path wildcard patterns, relative to any `INCLUDE_PATTERNS`,
```
junhaoliao:
```suggestion
  # @param {string[]} [EXCLUDE_PATTERNS=[]] Path wildcard patterns, relative to any `INCLUDE_PATTERNS`,
```

junhaoliao (changes requested):
for the title, how about:
```
fix(taskfiles): Output unexpected errors in checksum tasks; Call `gtar` on macOS (fixes #67); Add unit tests for checksum tasks.
```

exports/taskfiles/utils/checksum.yaml
```diff
@@ -25,58 +38,95 @@ tasks:
...
       - |-
         (
           {{- range .INCLUDE_PATTERNS}}
-          for path in {{.}}; do
-            test -e "$path"
-          done
+            for path in {{.}}; do
+              test -e "${path}" \
+                || (
+                  echo "Include path does not exist: ${path}" > "{{$.TMP_ERR_LOG}}"
```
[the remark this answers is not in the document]
junhaoliao: actually, i might be wrong about this one - we do `exit` right after an error is reported

junhaoliao (approved):

PR #91, opened 2025-10-13
fix(taskfiles)!: Prevent `CMAKE_SETTINGS_DIR` initialization race conditions in concurrent dependency installs using run-once `make-or-clear-dir`.

Bill-hbrhbr:
# Description

This PR partially addresses #89 by fixing concurrency issues in `cmake:install-deps-and-generate-settings`, where multiple tasks could attempt to recreate `CMAKE_SETTINGS_DIR` simultaneously when dependency installation workflows ran in parallel.

`misc:make-or-clear-dir` is introduced to safely initialize directories. This helper runs only once per unique directory path, ensuring that no two workflows interfere with each other.

As a result, `cmake:install-deps-and-generate-settings` now executes safely in parallel and avoids overwriting progress from other dependency workflows.

# Checklist

* [X] The PR satisfies the [contribution guidelines][yscope-contrib-guidelines].
* [X] This is a breaking change and that has been indicated in the PR title, OR this isn't a
  breaking change.
* [X] Necessary docs have been updated, OR no docs need to be updated.

# Validation performed

* [X] Package tasks for the CLP repo completes successfully.

## Summary by CodeRabbit

* **Chores**
  * Added a reusable directory-management utility to handle creation and safe clearing of directories, replacing ad-hoc shell cleanup.
  * Switched build/setup tasks to run only when inputs change to avoid redundant runs and race conditions.
  * Exposed the single-run-per-change behaviour in the public task declarations for more predictable initialization and aggregation of generated settings.

### DIFF
--- a/exports/taskfiles/utils/checksum.yaml
+++ b/exports/taskfiles/utils/checksum.yaml
@@ -16,6 +16,7 @@ tasks:
     requires:
       vars: ["CHECKSUM_FILE", "INCLUDE_PATTERNS"]
     cmds:
+      - "mkdir -p '{{ dir .CHECKSUM_FILE }}'"
       # We explicitly set `--no-anchored` and `--wildcards` to make the inclusion behaviour match
       # the default exclusion behaviour.
       #
--- a/exports/taskfiles/utils/remote.yaml
+++ b/exports/taskfiles/utils/remote.yaml
@@ -11,14 +11,12 @@ tasks:
   #
   # @param {string} URL
   # @param {string} FILE_SHA256 Content hash to verify the downloaded file against.
-  # @param {string} [OUTPUT_FILE={{(base .URL)}}] Path where the file should be stored.
+  # @param {string} OUTPUT_FILE Path where the file should be stored.
   curl:
     internal: true
     label: "{{.TASK}}-{{.OUTPUT_FILE}}"
-    vars:
-      OUTPUT_FILE: "{{default (base .URL) .OUTPUT_FILE}}"
     requires:
-      vars: ["FILE_SHA256", "URL"]
+      vars: ["FILE_SHA256", "OUTPUT_FILE", "URL"]
     generates: ["{{.OUTPUT_FILE}}"]
     status:
       - |-


exports/taskfiles/utils/misc.yaml
```diff
@@ -41,3 +43,47 @@ tasks:
...
+    internal: true
+    label: "{{.TASK}}:{{.FILE_PATH}}"
+    vars:
+      PARENT_DIR: "{{ dir .FILE_PATH }}"
+    requires:
+      vars: ["FILE_PATH"]
+    run: "when_changed"
+    preconditions:
+      # This task runs once per unique file path, which is fingerprinted by the task label. To avoid
+      # possible label collisions, we avoid using relative paths.
```
davidlion: Similar comments about doc strings and the name as above.

In this case the name is more serious. If the file doesn't already exist, it will not be created (so it won't be "reset", it just still won't exist) and potentially any parent directories will be created.

exports/taskfiles/utils/misc.yaml
```diff
@@ -41,3 +43,47 @@ tasks:
...
         pip3 install --upgrade pip
         pip3 install --upgrade -r "{{.REQUIREMENTS_FILE}}"
+
+  reset-dir:
+    desc: "Clears the directory at the given path without removing the directory itself."
+    internal: true
+    label: "{{.TASK}}:{{.DIR_PATH}}"
+    requires:
+      vars: ["DIR_PATH"]
+    run: "when_changed"
```
davidlion: If `run:once` produces the same behaviour (because we set `label`), let's use `run:once` as it is easier to read/understand.

exports/taskfiles/utils/misc.yaml
```diff
@@ -41,3 +41,23 @@ tasks:
...
+    preconditions:
+      - sh: "{{osIsAbs .DIR_PATH}}"
+        msg: "DIR_PATH `{{.DIR_PATH}}` is not absolute."
+    cmds:
+      - "rm -rf '{{.DIR_PATH}}'"
+      - "mkdir -p '{{.DIR_PATH}}'"
```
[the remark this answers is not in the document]
Bill-hbrhbr: fix your memory. Run "once" will only run the task once per task invocation regardless of the label. Run "when_changed" will run once per unique label/input var set.

PR #117, opened 2026-06-30
fix(taskfiles): Improve concurrency control for external tools:

davidlion:
- Default to the number of logical processors.
- Allow users to control the default through the env var `TASK_EXTERNAL_TOOL_CONCURRENCY`.

# Description

Currently, concurrency variables are not handled consistently so a user cannot easily control the concurrency from the command line. The goal of this PR is to create a consistent flow for users to easily control concurrency in resource limited scenarios.

For example, most tasks use the variable `JOBS`, but cmake also has a task that uses `CMAKE_JOBS` and passes that value as a `JOBS` argument. If the user sets `JOBS` it will not be respected by the task using `CMAKE_JOBS`. This is further complicated as top-level projects (e.g. CLP) end up using their own ad-hoc variables to control concurrency.

Tasks that set a concurrency related parameter for a tool should (see PR code for example):
1. Use the argument provided by the caller (if non-empty).
2. Use the environment variable `TASK_EXTERNAL_TOOL_CONCURRENCY` (if non-empty).
3. Fallback to the number of logical cores provided by `G_NPROCS` in misc.yaml.

Tasks that propagate a concurrency related parameter should:
1. Propagate it without explicitly setting the value unless they need to always override the user.
   * e.g. if the task `FOO` takes a parameter `JOBS_FOO` and calls `BAR` which takes `JOBS_BAR` it should just directly pass the arg `JOBS_BAR: {{.JOBS_FOO}}`.

A user wishing to set the default concurrency value can either set `TASK_EXTERNAL_TOOL_CONCURRENCY` in their environment, or at the commandline: `TASK_EXTERNAL_TOOL_CONCURRENCY=16 task ...`

# Checklist

  breaking change.

# Validation performed

Manually tested with the boost test and prints.

## Summary by CodeRabbit

* **New Features**
  * Build tasks now automatically use all available CPU cores by default when no parallelism value is provided.
  * Parallel build settings can now inherit from a shared default across related workflows.

* **Bug Fixes**
  * Improved consistency in how build concurrency is handled, reducing cases where builds fell back to tool-specific defaults.
  * Updated related task defaults so build and install steps behave more predictably together.

### DIFF
--- a/exports/taskfiles/utils/boost.yaml
+++ b/exports/taskfiles/utils/boost.yaml
@@ -62,8 +62,8 @@ tasks:
   # @param {string} BUILD_DIR Directory in which to build boost.
   # @param {string} INSTALL_PREFIX Path prefix of where the project should be installed.
   # @param {string[]} [EXTRA_ARGS] Any additional arguments to pass to the build command.
-  # @param {int} [JOBS={{default .G_NPROCS (env "TASK_JOBS")}}] The maximum number of concurrent
-  # processes to use when building.
+  # @param {int} [JOBS={{default .G_NPROCS (env "TASK_EXTERNAL_TOOL_CONCURRENCY")}}] The maximum
+  # number of concurrent processes to use when building.
   # @param {string} [CMAKE_SETTINGS_DIR] If set, the directory where the project's CMake settings
   # file should be stored.
   build-and-install:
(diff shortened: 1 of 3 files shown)


kirkrodrigues (approved):
For the PR title & body, how about:

```
fix(taskfiles): Improve concurrency control for external tools:
- Default to the number of logical processors.
- Allow users to control the default through the env var `TASK_EXTERNAL_TOOL_CONCURRENCY`.
```

